PR #11, opened 2020-05-19
Only annotate

hoelzer:
I want to run the nextflow w/o the virus detection part. Let's say I have a FASTA with viruses sequences that I just want to annotate and not filter for putative virus signals. 

This is now implemented in the nextflow via the flag

```
--onlyannotate
```

If the flag is set, the detection sub-workflow will be skipped. Renaming and Length filtering will be still applied. 

I also added `--viphog_version` and `--meta_version` because Guillermo and I are currently testing the different model versions we have. With the flags, I can more easily switch between the different tables (this is more a development thing)

I thought you might be interested in the changes, so please have a look, and if you are fine please feel free to merge this into master.

### DIFF
--- a/virify.nf
+++ b/virify.nf
@@ -31,7 +31,8 @@ if (workflow.profile == 'standard') {
   println "Output dir name: $params.output\u001B[0m"
   println " "
 }
-println "\033[2mDev ViPhOG database: $params.version\u001B[0m"
+println "\033[2mDev ViPhOG database: $params.viphog_version\u001B[0m"
+println "\033[2mDev Meta database: $params.meta_version\u001B[0m"
 println " "
         
 if( !nextflow.version.matches('20.01+') ) {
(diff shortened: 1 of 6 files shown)


virify.nf
```diff
@@ -549,12 +582,16 @@ def helpMSG() {
...
-                      v1: no additional bit score filter (--cut_ga not applied, just e-value filtered)
-                      v2: --cut_ga, min score used as sequence-specific GA, 3 bit trimmed for domain-specific GA
-                      v3: --cut_ga, like v2 but seq-specific GA trimmed by 3 bits if second best score is 'nan'
+    --viphog_version    define the ViPhOG db version to be used [default: $params.viphog_version]
+                        v1: no additional bit score filter (--cut_ga not applied, just e-value filtered)
+                        v2: --cut_ga, min score used as sequence-specific GA, 3 bit trimmed for domain-specific GA
+                        v3: --cut_ga, like v2 but seq-specific GA trimmed by 3 bits if second best score is 'nan'
+    --meta_version      define the metadata table version to be used [default: $params.meta_version]
+                        v1: older version of the meta data table using an outdated NCBI virus taxonomy 
+                        v2: 2020 version of NCBI virus taxonomy
```
mberacochea, line 594: Do we need the old version of the NCBI virus taxonomy?.

hoelzer: This is exactly the pain point Guillermo and I are working on. We currently achieve more taxon assignments with the metadata table v1 based on the old NCBI virus taxonomy, although the v2 is more complete. We investigate, if some of the additional assignments with v1 are just false positives. And yes, the idea is to simply use the updated v2 of this table. 

For now, it's easier for me to test v1 vs v2 with this flag :)

hoelzer:
> Looks good, I'll give this a try.

So I can merge this into master? (I don't want to destroy anything ;) )

virify.nf
mberacochea: Cool :)

mberacochea:
> > Looks good, I'll give this a try.
> 
> So I can merge this into master? (I don't want to destroy anything ;) )

Go ahead and merge it.

Issue #13, opened 2020-05-21
Mashmap step for the nextflow implementation


mberacochea:

mberacochea:
Completed in https://github.com/EBI-Metagenomics/emg-viral-pipeline/commit/9a76203f0146fdc48b5c0b6bcbe9b4ecc0772137

Issue #12, opened 2020-05-19
Replace Additonal_metadata.{version}.xls with a .tsv


mberacochea:
We can feed the file directly to ratio_evalue_table.py

mberacochea:
Fixed in https://github.com/EBI-Metagenomics/emg-viral-pipeline/commit/18d079945f9024edcd2f02299c536f959b1bd1e8#diff-6a5f79713563fa3f2cb7a10f9c5ba9c4
Pending: rename file in FTP server

Amend Nextflow implementation

mberacochea:
Completed in commit: https://github.com/EBI-Metagenomics/emg-viral-pipeline/commit/9a76203f0146fdc48b5c0b6bcbe9b4ecc0772137

PR #14, opened 2020-05-22
Merge DEV

mberacochea:
This PR has:
- CWL code, moved to an dev version in order to use conditionals.
- Changes for https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/13
- Changes for https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/12
- Fix for lenght filtering script, memory problem reported by Alex A.
- Mashmap optional step in CWL and Nextflow
- Travis build for CWL and python scripts. Not sure how to write tests for Nextflow, at least a validation of the code (I'll dig into this).

I've tested this locally and it's working, I'll move the code to the cluster to run some tests.


nextflow/modules/mashmap.nf
```diff
@@ -0,0 +1,19 @@
...
+      tuple val(assembly_name), val(confidence_set_name), file(fasta) 
+      file(reference)
+      val(length) 
+    
+    output:
+      file("${confidence_set_name}_mashmap_hits.tsv")
+    
+    shell:
+    """
+    mashmap -q ${fasta} -r ${reference} -t ${params.cores} -o ${confidence_set_name}_mashmap_hits.tsv --noSplit -s ${length}
```
hoelzer, line 17: * you can directly use `params.mashmap_len` here, since nf v20 params are global and can be simply used in modules

hint:
* the implementation works for input that should be compared to the same `reference` (and most likely this will be the case in one run) but maybe for a future extension and high-throughput processing let's assume you have two assemblies A and B and you want to mashmap them against reference refA and refB  

Now it's unclear for nextflow which reference belongs to which assembly. This can be resolved by the `--list` input:

```bash
#assemblies.csv
A,a.fasta
B,b.fasta

#mashmap.csv
A,refA.fasta
B,refB.fasta
```

And now instead of 

```java
            mashmap(predicted_contigs, mashmap_ref_ch)
```

one can do

```java
            mashmap(predicted_contigs.join(mashmap_ref_ch, by: 0))
```

and some other slight changes would need to be introduced. But by doing so, the correct references could be assigned to the correct assemblies. 

Totally fine like it is implemented now, I just wanted to mention this :)

hoelzer: ah and use `task.cpus` instead of `params.cores`. 

`params.cores` refers to the parameter in the `nextflow.config` file. 

but let's say you run the pipeline in the cloud where you defined in the corresponding config `cores '20'` for mashmap, this will not be used because you access this value via `task.cpus`. 

that's why you wrote 

```
withLabel: mashmap {      cpus = params.cores;   container = 'quay.io/biocontainers/mashmap:2.0--gsl2.2_1' }
```

`nextflow run virify.nf --cores 4` will set `params.cores = 4` and then `task.cpus = 4`

hoelzer:
Okay, that was annoying. In the mashmap module I changed

```
-s ${length}
```

to

```
-s ${params.mashmap_len}
```

because you can directly use parameters in modules. But it was not working (it was always using the default of 2000). 

I figured out that the problem seems to be if you are using 

```
shell:
"""
-s ${params.mashmap_len}
"""
```

or 
```
script:
"""
-s ${params.mashmap_len}
"""
```

In a `script:` environment `params.mashmap_len` is updated correctly when I change `--mashmap_len 4500` but not in the `shell:` environment! 

I will also check for other modules where `shell:` instead of `script:` is used and update this. 

I don't know if this is intended or a super annoying bug.

hoelzer:
tldr: 

either use 

```
shell:
'''
'''
```

or

```
script:
"""
"""
```

it seems that 

```
shell:
"""
"""
```

behaves strange

PR #15, opened 2020-05-27
10% first threshold and modified proportion per rank

hoelzer:
@guille0387 implemented discussed changes of the voting system to better reflect the updated NCBI taxonomy metadata file and general taxonomic assignments:

* lowered the 20% cutoff of discovered informative HMMs to 10%
* taxonomy is now assigned by comparing the number of HMMs from the same rank to the total number of identified models for this rank {genus/subfamily/family/order} 
    * for example: before the number of HMMs identified for a specific genus were compared to the total number of __all__ identified proteins and due to that the cutoff of 60% was not reached in many cases


hoelzer:
@guille0387 does this also implement the discussed minimum number (e.g. n = 2) of assigned models per rank?

hoelzer:
@mberacochea I added the evalue cutoff from the ratio script as an adjustable parameter to the nextflow implementation. I set a default value so the CWL should not be affected

hoelzer:
@mberacochea and I also parameterized the assign step in the nextflow implementation. 

@guille0387 so in the nextflow you can now use 

```
--taxthres 0.3 --prop 0.08
```

for example. Defaults are currently 

```
--taxthres 0.6 --prop 0.1
```

mberacochea:
Looks good, excellent job folks!

PR #18, opened 2020-06-02
switch to https://schema.org/version/latest/schema.rdf

mr-c:
You are invited to move/donate your tool descriptions to https://github.com/common-workflow-library/bio-cwl-tools where we do this sort of maintenance collaboratively


mberacochea:
Thanks @mr-c !

PR #22, opened 2020-06-04
Dev

hoelzer:
* implemented merged profile structure for Nextflow
    * now we have only four config files:
        * `container.config` --> has the docker information
        * `conda.config` --> has the conda information (unfortunately no PPR-Meta conda recipe)
        * `node.config` --> has information on CPU and RAM for HPC/Cloud execution
        * `local.config` --> has CPU information for local execution

* `--databases` can now be used to point to a directory to store all databases, also during local execution


nextflow.config
```diff
@@ -47,7 +47,7 @@ params {
 
     // development
     viphog_version = 'v3'
-    meta_version = 'v1'
+    meta_version = 'v2'
```
hoelzer, line 50: @mberacochea does the CWL pipeline now also use the v2 of the metadata file as the default? I think with the changes to the voting system we introduced it is fine now to switch to v2 as the default. v1 we might only need to reproduce some previous results

mberacochea: I was using v1, I'll change to v2 now :+1:

nextflow.config
```diff
@@ -61,76 +61,191 @@ params {
...
+                enabled = true
+                autoMounts = true
+                cacheDir = params.cachedir
+        }
+        includeConfig 'nextflow/configs/container.config' 
+    }
     
     ebi {
-        params.cloudProcess = true
   	    params.workdir = "/hps/nobackup2/production/metagenomics/$USER/nextflow-work-$USER"
```
mberacochea, line 157: In the cluster it's faster to use the /scratch area for temporary files.

nextflow.config
```diff
@@ -61,76 +61,191 @@ params {
...
+                enabled = true
+                autoMounts = true
+                cacheDir = params.cachedir
+        }
+        includeConfig 'nextflow/configs/container.config' 
+    }
 
     yoda {
-        params.cloudProcess = true
   	    params.workdir = "/hps/nobackup2/metagenomics/$USER/nextflow-work-$USER"
```
mberacochea, line 179: Same in Yoda (/scratch area is faster).

nextflow.config
hoelzer: alright! and good, then we have the same status for nf and cwl. the v2 has updated viral taxonomy names

nextflow.config
hoelzer: ah okay! so it would be good to set the default here to 

`/scratch/nextflow-work-$USER` ? 

how is the /scratch area cleaned? Is there something like every week? 

In any case, I agree it's good to point per default to the /scratch area for the tmp files and if one really needs to resume a run in half a year it's still possible to use `--workdir` to point to another area

mberacochea: The scratch are is cleaned every week.
In EBI the systems team mark that as a good practice, but being able to specify the folder is a good idea.

PR #33, opened 2020-10-22
Voting

hoelzer:
* Guillermo added some small changes to the voting system
    * now, if a taxonomy is assigned on genus level, the full lineage is assigned and no further checks are performed for family, etc...


hoelzer:
we need to fix something, I cancel this PR

PR #41, opened 2021-01-12
Dev

mberacochea:
@hoelzer if this looks OK we can then release a new version :+1:


docker/r_fungi/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM rocker/r-ver:3.5.0
```
hoelzer, line 1: This Dockerfile is used for the ChromoMap ploting and I hijacked it from another source where it was used in context of a fungi annotation project :) Thus, we could also rename it to something like `r_chromomap` to be more descriptive

hoelzer: and: 

```docker: Error response from daemon: pull access denied for microbiomeinformatics/r_fungi, repository does not exist```

hoelzer:
@mberacochea some images are not yet on your DockerHub account, right? e.g.

- fungi_r (that we might rename, see comment above)
- blast
- ...

hoelzer (approved):
@mberacochea works and looks good on my end!

Just some functionalities are not working atm because the containers are not yet on your DockerHub account. 

I also enabled Nextflow DSL2 now bc/ it's stable and widely used now (thus suppressing a warning that occurred before). 

I think after that merge we should have a release 0.3.0 ?

docker/r_fungi/Dockerfile
mberacochea: good catch, I'll fix this.

mberacochea:
> @mberacochea works and looks good on my end!
> 
> Just some functionalities are not working atm because the containers are not yet on your DockerHub account.
> 
> I also enabled Nextflow DSL2 now bc/ it's stable and widely used now (thus suppressing a warning that occurred before).
> 
> I think after that merge we should have a release 0.3.0 ?

I'll review the containers, I push those that are missing (most of them should be there, i.e. blast https://hub.docker.com/repository/docker/microbiomeinformatics/blast).

After checking that I'll merge the code and create the release 0.3.0 !

hoelzer:
@mberacochea wouhuh! agree! ;)

PR #48, opened 2021-04-12
Logo

hoelzer:
I just had the idea of generating a logo for VIRify and added it to the repository and the README ;) 

What do you think?


mberacochea:
Superb, I'll merge this but I need to fix the build first :(

PR #54, opened 2021-06-18
Dev martinh

hoelzer:
@mberacochea I updated the README a bit and commented the nf code. Also I checked that the nf pipeline runs through. So just some minor updates. Would be cool if we can do another release then!


mberacochea (approved):
Looks good to me :)

PR #56, opened 2021-07-02
Virify - MGnify protein MGYP assign step (CWL).

mberacochea:
Optional step in the CWL version. We have included
an additional input file with the assembly protein
accessions (MGYPs).
This fill will be used to replace the prodigal predicted
proteins with the corresponding MGYP.


KeteSakharova (approved):
Looks fine for me! 
I double checked that function that generates digest in pipeline-v5 corresponds to create_digest in viral pipeline. 
It is important because protein DB uses a bit another schema: it does digest for partial+seq, but pipeline-v5 only for seq.

caballero (approved):
code modifications are ok

PR #59, opened 2021-10-18
deactivate tower nf extension and simplify nf version check

hoelzer:
* deactivate the nf tower plugin
    * causes problems on systems that have restricted internet access
* simplify nf version check via nextflow.config


hoelzer:
@mberacochea these are just some small tweaks to get VIRify in the nf version also working on systems that have a more restricted access to public URLs. Should be save to merge

PR #75, opened 2022-09-17
Pprmeta repo fix

### LINKED ISSUE #69: Update PPR-Meta install
Hi,

I'm trying to install and run the pipeline using Singularity on an HPC. The install fails during the PRR-Meta installation process giving me the following error:

```
Error executing process > 'download_pprmeta:pprmetaGet'

Caused by:
  Process `download_pprmeta:pprmetaGet` terminated with an error exit status (128)

Command executed:

  git clone https://github.com/Stormrider935/PPR-Meta.git
  mv PPR-Meta/* .  
  rm -r PPR-Meta

Command exit status:
  128

Command output:
  Cloning into 'PPR-Meta'...

Command error:
  fatal: unable to look up current user in the passwd file: no such user
  Unexpected end of command stream
```

A quick look at github shows that the current git repository for PRR-Meta is https://github.com/mult1fractal/PPR-Meta not Stormrider935.

Are there any possible workarounds you would suggest? Thanks a lot for the pipeline!

hoelzer:
Should fix #69


hoelzer:
@mberacochea I did some clean-up here bc/ actually it makes not much sense to have such specific profiles like for `yoda` in the repo. If someone wants to use a specific nextflow profile this can be injected locally. Also I thought making the `cachedir` parameter in the nextflow pipeline more descriptive. 

Finally, I just corrected the URL from which PPR-Meta is cloned which should solve #69 . If there is no further feedback from that issue, I suggest to also just merge this PR. If there are no objections.

hoelzer:
Howver, I'm no also thinking if it's so smart to rename the `cachedir` variable into `singularity_cachedir` -- this will break pipelines that people might already running. I could add an error message to state that `cachedir` is deprecated and `singularity_cachedir` should be used in the future.

mberacochea:
Brilliant. I wouldn't worry to much about the cachedir.. let's just add that to the release notes.

I just adjusted the EBI profile a bit

mberacochea:
Ah, also I changed the base branch to dev

mberacochea (approved):

Issue #50, opened 2021-04-23
error running the pipeline with Nextflow version 20.10


uludag:
I get the following error when running the pipeline with Nextflow docker option

> Error executing process > 'preprocess:rename (1)'
> ...
> Command error:
>   Command 'ps' required by nextflow to collect task metrics cannot be found

I have checked that the relevant Docker image is missing the _ps_ command

> $ docker run --rm microbiomeinformatics/emg-viral-pipeline-python3:v1 ps
> docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "ps": executable file not found in $PATH: unknown.

--mahmut

hoelzer:
@uludag thanks for reporting!

Ah, this happened when we switched from a set of different image sources to harmonize everything in [one place](https://hub.docker.com/search?q=microbiomeinformatics&type=image). 

@mberacochea the containers all need `ps` installed to function properly with the in-build runtime/resource collection in `nextflow`, e.g. something like:

```docker
RUN apt update && apt install -y procps wget gzip && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```

Can you take care of that? Or can I simply adjust the Dockerfiles and then they get autobuild and pushed to your team Dockerhub? 

@uludag Did you start the pipeline with `-with-timeline` or `-with-report`? And from the `master` branch? If you use the latest release tag `v0.2.0` it should work because other images are used, for example please try:

```bash
nextflow run EBI-Metagenomics/emg-viral-pipeline -r v0.2.0 --fasta "/home/$USER/.nextflow/assets/EBI-Metagenomics/emg-viral-pipeline/nextflow/test/assembly.fasta" --cores 4 -profile local,docker
```

and adjust `-profile` according to your system.

mberacochea:
I see, Nextflow needs a the following utilites installed to be able to collect the metrics:
```Make sure the following tools are available awk, date, grep, egrep, ps, sed, tail, tee in the system where the jobs are executed``` from: https://www.nextflow.io/docs/latest/tracing.html.

I will include them on the containers, I can provide an ETA but I will it as soon as I can.

uludag:
@hoelzer  thanks for the example command-line, using the release tag `v0.2.0` I'm able to run the emg-viral workflow.

[...]

Issue #74, opened 2022-09-16
Upgrade Spades to the latest version


mberacochea:

hoelzer:
Good point, is there a new container @mberacochea ? Should be then easy to add to the `nextflow` pipeline version as well (I can also do that, as you prefer)

PR #80, opened 2022-11-25
CheckV - prophages fixes and post processing gff creation step

mberacochea:
This PR includes:
- A fix for prophages and chechV
- GFF generation step

PR #81, opened 2022-11-26
 Update containers, nf trace compatible now. Some to biocontainers.

mberacochea:
This branch has a the following:
- Moved over to quay.io (dockerhub has pull limits now)
- Migrated some tools to biocontainers, less maintaince work
- Added the tools required for Nextflow monitoring and trace

TODO: complete the build.py command to push all the containers.
I would like to use gh actions to push the containers when
we get a new release.

NOTE: it was built on top of https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/80

Should close https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/50


containers/README.md
```diff
@@ -1,4 +1,14 @@
-# Docker files for EMG Viral Pipeline
+# Containers for VIRify
+
+## Container registry
+
+The containers used on this pipeline are stored under the [Microbiome Informatics]() (Quay organization)[https://quay.io/repository/microbiome-informatics/].
+
+## Build process
+
+The build process takes care of tagging the images using the LABELS on the Dockerfiles of the tools.
```
hoelzer, line 9: yeah, as said in the other PR you could use the LABEL + md5 sum of the Dockerfile for tagging. Then you can have multiple containers with the same tool version but also with other changes (added dependencies, ...)

hoelzer (approved):
Nice! +1

Ad automation container build

> TODO: complete the build.py command to push all the containers. I would like to use gh actions to push the containers when we get a new release.

That would be great. Would you like to set it up so that all the Dockerfiles are checked, and when there is a VERSION change (or different checksum), the container(s) get auto-build and pushed? This should work via gh Action somehow. But so far I only did this via having a single repository per Dockerfile. And then, when this Dockerfile has been updated the gh Action builds a new container and does the push.

mberacochea:
This one also closes #74

containers/README.md
mberacochea: Yeah... that is a good idea -> https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/85

Issue #93, opened 2023-03-01
Regression - VirSorter prophage report at the end of the contigs


mberacochea:
By updating the virsorter container to the one hosted in biocontainers we introduced a regression: https://github.com/simroux/VirSorter/issues/68.
VirSorter extends the prophage sequence by 50 nucleotide beyond the last gene in 5' and 3' (to include potential att sites and not end a contig right on a start / stop codon). I just forgot to include a check to make sure we don't extend past the contig, i.e. if the last gene of the prophage is at the end of the contig, the coordinate from VirSorter will be 50 nucleotides beyond the contig (hence 63493 vs 63443).

mberacochea:
Fixed by https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/94

PR #94, opened 2023-03-01
Re-build the virsorter container, using one hosted in quay by our team.

mberacochea:
Added 2 more cwl test for virsorter to cover this case and to avoid regressions in the future.

We need to port the tests to nextflow.. not sure how yet.

This PR will fix https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/93


mberacochea:
> Very cool! It looks like no test for nextflow, but cwl probably enough :)

Yes, we don't have unit tests for the nextflow code (_yet_). It was easier to improve the cwl tests for that tool

mberacochea:
@hoelzer I'm merging this one as we need. I'll tag you in the dev -> master PR, we will add other bits to that one.

Issue #92, opened 2023-02-21
Missing CheckV output when there are multiple prophages in contig


Ales-ibt:
Hello there!

I'm running Virify through Nextflow and realized the CheckV directory is missing when more than one prophage is predicted in the same contig. Tracing back the log files, I found the error:

`Please remove duplicated sequence IDs from the input FASTA file: XXX: cannot stat 'prophages/quality_summary.tsv': No such file or directory
`

Adding the coordinates to the sequence IDs in the provirus fasta file would solve the problem.

Bests!

mberacochea:
Perfect, thank you Ale. I'll take a look

Ales-ibt:
Me again...

After taking a look at the output files and tracing back the checkV results, I have the following summary of problems associated with the integration of the outputs:

**Problems related to the integration of predicted phages and quality:**
1. CheckV is running on the full contig for prophages 
2. Multiple phages occurring in the same contig are missing
3. Prophages coordinates in the gff don't correspond to coordinates prediction
4. Prophages with no CDS matching ViPhOG are excluded in the gff (is this for quality control?)
5. Viral genomes in full contigs are missing in the gff output

**Problems related to the GFF file format (which is not validated):**
1. Protein IDs in gff file are missing
2. Add the mobile_genetic_element feature and the corresponding coordinates. Then link the corresponding CDS using the 'Parent' key in the attributes field
3. Fix the CDS coordinates in the context of the full contig instead of the subsequence

Find attached some figures I prepared for this report.

Bests,


Alejandra.
[virify_bug_Feb2023.pdf](https://github.com/EBI-Metagenomics/emg-viral-pipeline/files/10804601/virify_bug_Feb2023.pdf)

hoelzer:
Hey @Ales-ibt  - wow, thanks for the detailed report and visuals!

mberacochea:
Fixed by https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/95 and https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/94

PR #95, opened 2023-03-09
Bugfix/issue 92 prophages and gff generation

Ales-ibt:
Hello!

Please review and merge to dev. 

Changes in this branch fix issue [#92](https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/92)

Bests.

### DIFF
--- a/bin/write_viral_gff.py
+++ b/bin/write_viral_gff.py
@@ -67,7 +67,7 @@ def collect_annot(virify_files, sample_prefix, ena_mapping=None):
                 if not Best_hit == 'No hit':
                     Best_hit=Best_hit.replace('.faa','')
                     Label=line_l[7]
-                    annotation='viphog='+Best_hit+';'+'viphog_taxonomy=\''+Label+'\''
+                    annotation='viphog='+Best_hit+';'+'viphog_taxonomy='+Label
                 #else:
                 #    annotation='product=hypothetical_protein'
 
@@ -125,7 +125,7 @@ def write_gff(
                         except:
                             lineage.append(element)
                 if len(lineage)>0:
-                    taxo_string=';'.join(lineage)
+                    taxo_string='%3B'.join(lineage)
                 else:
                     taxo_string='unclassified'
 
@@ -183,7 +183,7 @@ def write_gff(
                 region_end=contigs_len_dict[element]
             
             if element in taxonomy_dict.keys():
-                attributes='ID='+element_id+';gbkey=mobile_element;mobile_element_type='+contig_data[element]+';'+checkv_dict[element]+';taxonomy=\''+taxonomy_dict[element]+'\''
+                attributes='ID='+element_id+';gbkey=mobile_element;mobile_element_type='+contig_data[element]+';'+checkv_dict[element]+';taxonomy='+taxonomy_dict[element]
             else:
                 attributes='ID='+element_id+';gbkey=mobile_element;mobile_element_type='+contig_data[element]+';'+checkv_dict[element]
 
--- a/bin/write_viral_gff.py
+++ b/bin/write_viral_gff.py
@@ -21,136 +21,213 @@ def get_ena_contig_mapping(ena_contig_file):
     return ena_mapping
 
 
-def write_gff(virify_files, sample_prefix, ena_mapping=None):
-    contigs = {}
+def collect_annot(virify_files, sample_prefix, ena_mapping=None):
+    ## Parsing the annotation files
+    contig_protID={}
+    prot_data={}
+    contig_data={}
+    for vfile in virify_files:
+        with open(vfile,'r') as input_table:
+            next(input_table)
+            for line in input_table:
+                line_l=line.rstrip().split('\t')
+                Contig=line_l[0]
+                CDS_ID=line_l[1]
+                Start=line_l[2]
+                if Start == '0':
+                    Start=1
+                End=line_l[3]
+                Direction=line_l[4]
+                Best_hit=line_l[5]
+
+                if '|phage-circular' in Contig:
+                    Contig=Contig.replace('|phage-circular','')
+                    CDS_ID=CDS_ID.replace('|phage-circular','')
+                    desc='phage_circular'
+
+                elif '|prophage-' in Contig:
+                    desc='prophage'
+                    # Fixing CDS coordinates to the context of the whole contig
+                    prophage_start=int(Contig.split('|')[1].split('-')[1].split(':')[0])
+                    Start=int(Start)+prophage_start
+                    End=int(End)+prophage_start
+
+                else:
+                    Contig=Contig.replace('|','')
+                    CDS_ID=CDS_ID.replace('|','')
+                    desc='phage_linear'
+
+                if Contig not in contig_protID.keys():
+                    contig_protID[Contig]=[CDS_ID]
+                    contig_data[Contig]=desc
+                else:
+                    contig_protID[Contig].append(CDS_ID)
+
+                Direction=Direction.replace('-1','-').replace('1','+')
+
+                if not Best_hit == 'No hit':
+                    Best_hit=Best_hit.replace('.faa','')
+                    Label=line_l[7]
+                    annotation='viphog='+Best_hit+';'+'viphog_taxonomy=\''+Label+'\''
+                #else:
+                #    annotation='product=hypothetical_protein'
+
+                    values=(str(Start),str(End),Direction,annotation)
+                    prot_data[CDS_ID]=values
+
+    #return dictionaries
+    return contig_protID, prot_data, contig_data
+
+
+def write_gff(
+    checkv_files, taxonomy_files, sample_prefix, assembly_file, annot_dicts, ena_mapping=None
+):
+
     if ena_mapping:
         ena_assembly_accession = list(ena_mapping.values())[0].split(".")[0]
         output_filename = f"{ena_assembly_accession}_virify.gff"
     else:
         output_filename = f"{sample_prefix}_virify.gff"
-    with open(output_filename, "w") as gff:
-        gff.write(f"##gff-version 3\n")
-        #   annotation type specify this is virify confidence level
-        for vfile in virify_files:
-            virify_df = pd.read_csv(vfile, sep="\t")
-            filtered_df = virify_df[virify_df["Best_hit"] != "No hit"]
-            #   set lowest possible start and highest possible end
-            for index, row in filtered_df.iterrows():
-                if not row["Contig"] in contigs:
-                    contigs[row["Contig"]] = {"start": row["Start"], "end": row["End"]}
-                else:
-                    if contigs[row["Contig"]]["start"] > row["Start"]:
-                        contigs[row["Contig"]]["start"] = row["Start"]
-                    if contigs[row["Contig"]]["end"] < row["End"]:
-                        contigs[row["Contig"]]["end"] = row["End"]
 
-                #   remove '.faa' trailing viphog ID
-                viphog = row["Best_hit"].strip(".faa")
-                direction = "-" if row["Direction"] == "-1" else "+"
-
-                #   change contig name if ena mapping required
-                if ena_mapping:
-                    contig_name = f'{ena_mapping[row["Contig"]]}-{row["Contig"]}'
-                else:
-                    contig_name = row["Contig"]
-
-                #   ID=ERZ2271866.1-NODE-1-length-21396-cov-5.122534;viphog=ViPhOG1;viphog_taxonomy=Phaeovirus
-                annotation = (
-                    f'ID={contig_name};viphog={viphog};viphog_taxonomy={row["Label"]}'
-                )
-                #   start with: ERZ2271866.1-NODE-1-length-21396-cov-5.122534	ViPhOG	proviral_region	1020	2050	.	-	.
-                gff.write(
-                    f'{contig_name}\t{viphog}\tviral_sequence\t{row["Start"]}\t{row["End"]}\t.\t{direction}'
-                    f"\t.\t{annotation}\n"
-                )
-    return contigs
-
-
-def write_metadata(
-    checkv_files, taxonomy_files, sample_prefix, virify_contigs, ena_mapping=None
-):
-    if ena_mapping:
-        ena_assembly_accession = list(ena_mapping.values())[0].split(".")[0]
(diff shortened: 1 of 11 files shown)


hoelzer:
I am running some large-scale tests now on a HPC w/ SLURM

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
-            #   set lowest possible start and highest possible end
-            for index, row in filtered_df.iterrows():
-                if not row["Contig"] in contigs:
-                    contigs[row["Contig"]] = {"start": row["Start"], "end": row["End"]}
+def collect_annot(virify_files, sample_prefix, ena_mapping=None):
+    ## Parsing the annotation files
+    contig_protID={}
+    prot_data={}
+    contig_data={}
+    for vfile in virify_files:
```
mberacochea, line 28: Can you please run `black` to format the file?

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
-                    f"\t.\t{annotation}\n"
-                )
-    return contigs
-
-
-def write_metadata(
-    checkv_files, taxonomy_files, sample_prefix, virify_contigs, ena_mapping=None
+                    contig_protID[Contig].append(CDS_ID)
+
+                Direction=Direction.replace('-1','-').replace('1','+')
```
mberacochea, line 65: We use `snake_case` as the naming convention. Variables should not be un uppercase, `Direction` should be `direction`

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+
+                if not Best_hit == 'No hit':
+                    Best_hit=Best_hit.replace('.faa','')
+                    Label=line_l[7]
+                    annotation='viphog='+Best_hit+';'+'viphog_taxonomy='+Label
+                #else:
```
mberacochea, line 71: Please remove remove the commented code

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
-        output_filename = f"{ena_assembly_accession}_virify_contig_viewer_metadata.tsv"
+        output_filename = f"{ena_assembly_accession}_virify.gff"
     else:
-        output_filename = f"{sample_prefix}_virify_contig_viewer_metadata.tsv"
-    headers = (
-        "sequence_id\tcontig\tvirify_taxonomy\tstart_of_first_viphog\tend_of_last_viphog\tcheckv_provirus\t"
-        "checkv_quality\tmiuvig_quality\n"
-    )
-    checkv_dict, taxonomy_dict = {}, {}
+        output_filename = f"{sample_prefix}_virify.gff"
```
mberacochea, line 89: Don't we need the `.tsv`? Sorry, I'm confused

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+def collect_annot(virify_files, sample_prefix, ena_mapping=None):
+    ## Parsing the annotation files
+    contig_protID={}
+    prot_data={}
+    contig_data={}
+    for vfile in virify_files:
+        with open(vfile,'r') as input_table:
+            next(input_table)
+            for line in input_table:
+                line_l=line.rstrip().split('\t')
```
mberacochea, line 32: it's easier to read like this:
```suggestion
                contig, cds_id, start, end, direction, best_hit = line.rstrip().split('\t')
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+
+                if '|phage-circular' in Contig:
+                    Contig=Contig.replace('|phage-circular','')
+                    CDS_ID=CDS_ID.replace('|phage-circular','')
+                    desc='phage_circular'
+
+                elif '|prophage-' in Contig:
+                    desc='prophage'
+                    # Fixing CDS coordinates to the context of the whole contig
+                    prophage_start=int(Contig.split('|')[1].split('-')[1].split(':')[0])
```
mberacochea, line 50: Please add an example of a line here, it's going to help in the future to read this code
```python
# Fixing CDS coordinates to the context of the whole contig
# The contig name contains the positions in this format....
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
-                checkv_type = row["provirus"]
-                checkv_dict[row["contig_id"]] = {
-                    "checkv_type": checkv_type,
-                    "checkv_quality": row["checkv_quality"],
-                    "miuvig_quality": row["miuvig_quality"],
-                }
+        with open(cfile,'r') as input_table:
+            next(input_table)
+            for line in input_table:
+                line_l=line.rstrip().split('\t')
```
mberacochea, line 97: This is a bit hard to read, I would suggest to re-format like this:
```python
line_split = line.rstrip().split('\t')
contig_id = line_split[0].strip()
checkv_type = line_split[2].strip()
checkv_quality = line_split[7].strip()
miuvig_quality=  line_split[8].strip()

checkv_info = ";".join([
    f"checkv_provirus={checkv_type}"',
    f"checkv_quality={checkv_quality}",
    f"miuvig_quality={miuvig_quality}"
]) 
checkv_dict[contig_id] = checkv_info
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+                contig_id=line_l[0]
+                checkv_type=line_l[2]
+                checkv_quality=line_l[7]
+                miuvig_quality=line_l[8]
+                value='checkv_provirus='+checkv_type+';checkv_quality='+checkv_quality+';miuvig_quality='+miuvig_quality
+                checkv_dict[contig_id]=value
 
     #   parse taxonomic lineage when available
```
mberacochea, line 105: Can you improve the comment please, what is the output of this method as it's long and hard to understand.

An input and output in the comment would be enough to undernstand:
```
# Convert the taxonomy from XX to YY
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+        seq_len=len(str(record.seq))
+        contigs_len_dict[contig_id]=str(seq_len)
+
+    contig_protID=annot_dicts[0]
+    prot_data=annot_dicts[1]
+    contig_data=annot_dicts[2]
+    # Writing the gff file
+    with open(output_filename, "w") as gff:
+        gff.write("##gff-version 3\n")
+        contig_protID=annot_dicts[0]
```
mberacochea, line 146: These 3 lines are repeated

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+        prot_data=annot_dicts[1]
+        contig_data=annot_dicts[2]
+
+        # Writing the gff header
+        printed_contig=[]
+        for element in contig_protID.keys():
+            if 'prophage' in element:
```
mberacochea, line 153: there is no need to the else here:
```python
clean_id = element
if 'prophage' in element:
    clean_id = element.split('|')[0]
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+        phage_counter=0
+        phage_ids={}
+        for element in contig_protID.keys():
+            phage_counter+=1
+            element_id='phage_'+str(phage_counter)
+            phage_ids[element]=element_id
+            if 'prophage' in element:
+                seq_type='prophage'
+                seqid=element.split('|')[0]
+                start=element.split('|')[1].split('-')[1].split(':')[0]
```
mberacochea, line 173: These split calls are repeated in several places? Do you think we can write a method that returns the values instead if repeating the split calls?

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+                if int(region_end) > int(contigs_len_dict[seqid]):
+                    region_end=str(contigs_len_dict[seqid])
+            else:
+                seq_type='viral_sequence'
+                seqid=element
+                start='1'
+                region_end=contigs_len_dict[element]
+            
+            if element in taxonomy_dict.keys():
+                attributes='ID='+element_id+';gbkey=mobile_element;mobile_element_type='+contig_data[element]+';'+checkv_dict[element]+';taxonomy='+taxonomy_dict[element]
```
mberacochea, line 186: Improved readability:
```suggestion
        attributes = ";".join([
            f"ID={element_id}",
            f"gbkey={mobile_element}",
            f"mobile_element_type={contig_data[element]}",
            f"checkv_dict[element]'",
            f"taxonomy={taxonomy_dict[element]}"
        ])
```

bin/write_viral_gff.py
```diff
@@ -21,136 +20,213 @@ def get_ena_contig_mapping(ena_contig_file):
...
+                    source='Prodigal'
+                    seq_type='CDS'
+                    start=prot_data[protein][0]
+                    end=prot_data[protein][1]
+                    if int(end) > int(region_end):
+                        end=region_end
+                    score='.'
+                    strand=prot_data[protein][2]
+                    phase='0'
+                    #attributes='ID='+protein+';Parent='+element_id+';gbkey=CDS;'+prot_data[protein][3]
```
mberacochea, line 210: Please remove unused code

nextflow/modules/blast.nf
```diff
@@ -24,6 +25,7 @@ process blast {
     """
     else if (task.attempt.toString() == '2')
     """
+      sed -i "s/ /|/" ${fasta}
```
mberacochea, line 28: Sorry, I don't remember. Why do we need this sed instruction here?

nextflow/modules/prodigal.nf
```diff
@@ -13,6 +13,7 @@ process prodigal {
     
     script:
     """
+    sed -i "s/ /|/" ${fasta}
```
mberacochea, line 16: This one is repeated in several places, it would be better to do that in a single place.. in the rename step probably

mberacochea (changes requested):
Thank Ale! Great stuff, I've added some comments :)

nextflow/modules/blast.nf
hoelzer: I think only to have the complete FASTA ID information in the blast result. Because otherwise, blast would split the ID at the first occurring whitespace.

nextflow/modules/prodigal.nf
hoelzer: agree

bin/write_viral_gff.py
Ales-ibt: We don't need it anymore, checkv quality, taxonomy, and viphog annotation are now included in the attributes field of the gff output

bin/write_viral_gff.py
Ales-ibt: We cannot do it that way because the table has 6 fields when `best_hit=='No hit'` and 7 fields when `best _hit != 'No hit'`

hoelzer:
Large-scale test w/ a 1.3 GB FASTA and 2,255,782 DNA contigs run through and produced all output files. (SLURM, Docker).

I am also running some further tests.

nextflow/modules/blast.nf
Ales-ibt: Yes! We need it to ensure consistency across checkV and proteins prediction, thus resolving the issue when multiple prophages were present in the same contig.

nextflow/modules/prodigal.nf
Ales-ibt: Not sure when rename step is happening, but we need to rename the prophage sequence IDs after the viral predictions are done.

bin/write_viral_gff.py
Ales-ibt: I've changed all the variables starting with uppercase for lowercase

bin/write_viral_gff.py
Ales-ibt: Done!

bin/write_viral_gff.py
Ales-ibt: done!

bin/write_viral_gff.py
Ales-ibt: done

bin/write_viral_gff.py
Ales-ibt: done!

bin/write_viral_gff.py
Ales-ibt: I've removed repeated lines and re-formatted as suggested in the next comment

bin/write_viral_gff.py
Ales-ibt: agree

Ales-ibt:
I've run a small test with the changes you recommended and everything looks fine. 
The updated script can be found in the new commit.

bin/write_viral_gff.py
Ales-ibt: done!

bin/write_viral_gff.py
mberacochea: OK. We need to propagate this to any other pipelines that may be using this.
And we need to include this in the change log.

bin/write_viral_gff.py
mberacochea: Alright

nextflow/modules/blast.nf
mberacochea: I see. Then, it would be more convenient to use a different separator for our annotations.
Instead of:
```
>seq1 prophage-475020:518621
```
we should use:
```
>seq1_prophage-475020:518621
```
then we don't have to use sed and all the names will match everywhere

nextflow/modules/prodigal.nf
mberacochea: The problem comes from our parse_viral_predictions script... see my previuos comment

mberacochea (approved):
I think this one is ready to be merged

PR #98, opened 2023-04-13
Fasta IDs renaming fixed

Ales-ibt:
A couple of fixes in the scripts that were modifying the headers. Also there was an error in the gff-writer scripting the step of fixing the CDS coordinates


bin/write_viral_gff.py
```diff
@@ -225,6 +227,11 @@ def empty_if_number(string):
                     # Prophages include the start and the end in the string
                     # encoding: prophage:{prophage_start}-{prophage_end}
                     start, end = viral_seq_type.split("prophage-")[1].split(":")
+
+                    if int(start) == 0:
+                        start = '1'
+                        id_=id_.replace('prophage-0:','prophage-1:')
```
mberacochea, line 233: This only affects the first position? Sorry, I may be confused

mberacochea (changes requested):
Thank you @Ales-ibt . I've added some comments, I think we need to fix a different method of the rename script instead of adding that `flag` variable with the `|` character.

mberacochea (commented):
@Ales-ibt I've changed the code a bit... please give it a read, but I think we can merge this one.

mberacochea (approved):

PR #112, opened 2023-09-06
Fix the Chromomap container.

mberacochea:


mberacochea:
BTW we don't have to make a new release, as I've pushed the container to quay already.

hoelzer:
Nice, works! I will merge this directly and yes, we can skip the release bc the container has the same name.

hoelzer:
Ah, I think we still have to merge from `dev` into `master` now

hoelzer (commented):
+1

PR #122, opened 2024-04-25
remove lsf requirement for ebi profile

Vkale1:
Remove LSF for EBI profile will allow us to run with both ebi,slurm and ebi,lsf options.


KateSakharova (approved):
I think that should work

Issue #127, opened 2024-07-15
Question regarding GFF file generation


fischer-hub:
Hey, thanks for this nice pipeline!

I'm running the pipeline with the `--list` argument providing a sample sheet to it to run on  ~100 samples. But for all the viral sequences that are identified by the pipeline I only get one `GFF` file in the output directory of the first sample from the sample sheet. I think this is cause by [this](https://github.com/EBI-Metagenomics/emg-viral-pipeline/blob/master/virify.nf#L485)  line of code and I was just wondering what the reasoning behind taking only runnning the `GFF` script on the first contig file as a reference sequence is. 

I get that in the case of running the pipeline on a single sample this doesn't make a difference, but when dealing with multiple samples wouldn't it make sense to generate a GFF file for every sample where viral sequences have been found? Just wondering if I'm missing something here!

Thanks in advance!

mberacochea:
Hey @fischer-hub 

That looks like a bug to me, we usually run the pipeline one assembly at the time. So I think it's just an unhappy coincidence. I'll change it and will test the pipeline.

Cheers

fischer-hub:
Thanks again!

KateSakharova:
Hi, `--list` option would be removed in comming release. Multiple run would be supported using samplesheet and gathering samples should not be a problem anymore

PR #128, opened 2024-07-17
Add VirSorter2 process as an alternative to VirSorter

fischer-hub:
Added [VirSorter2](https://github.com/jiarong/VirSorter2) as a new process that can be used instead of `VirSorter` using the new flag `--use_virsort2`, adapted the downstream processes (the `parse` process and GFF generation script) to work with the different results of `VirSorter2`. E.g.: `VirSorter2` reports a confidence score (0-1) for every viral hit in the input data instead of a category.

Also had to add some changes to the GFF generation process because I always ran into file collision issues when more than one of the input samples reported significant viral sequences, since the `VirSort` and  `VirFinder` reports have the same file name for every sample. Might also be solved by running the GFF script for every sample containing viral seqs instead of once for all samples but I already asked about this in #127, maybe I'm missing something here!

### DIFF
--- a/virify.nf
+++ b/virify.nf
@@ -427,10 +427,9 @@ workflow detect {
 
         // parsing predictions
         parse(length_filtered_ch.join(virfinder.out).join(virsorter_out).join(pprmeta.out))
-        parse.out.view{ it -> "$it" }
 
     emit:
-        parse.out.join(renamed_ch).view().transpose().map{name, fasta, vs_meta, log, renamed_fasta, map -> tuple (name, fasta, map)}
+        parse.out.join(renamed_ch).transpose().map{name, fasta, vs_meta, log, renamed_fasta, map -> tuple (name, fasta, map)}
 }
 
 
--- a/virify.nf
+++ b/virify.nf
@@ -504,9 +504,9 @@ workflow annotate {
         def k = 1 
 
         // enumerate collected files to prevent file collision
-        viphos_annotations = annotation.out.map { _, __, annotations -> tuple(annotations, i++) }.collect(){it -> it[0]} //{ annotations, count -> "$annotations".replace('.', '_' + count + '.') }
-        taxonomy_annotations = assign.out.map { _, __, taxonomy -> tuple(taxonomy, j++) }.collect(){it -> it[0]} //{ taxonomy, count -> "$taxonomy".replace('.', '_' + count + '.') }
-        checkv_results = checkV.out.map { _, __, quality_summary, ___ -> tuple(quality_summary, k++) }.collect(){it -> it[0]} //{ quality_summary, count -> "$quality_summary".replace('.', '_' + count + '.') }
+        viphos_annotations = annotation.out.map { _, __, annotations -> tuple(annotations, i++) }.collect(){it -> it[0]}
+        taxonomy_annotations = assign.out.map { _, __, taxonomy -> tuple(taxonomy, j++) }.collect(){it -> it[0]}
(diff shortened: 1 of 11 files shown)


hoelzer:
Great, thanks a lot!

@mberacochea @guille0387: @fischer-hub is working with me at the RKI and we anyway wanted to use VS2 for some annotations. Thus, David kindly added the functionality to VIRify. Of couse, this would need more proper benchmarking but the idea would be to swap VS against VS2 at some point. 

For now, there would be this parameter switch possible.

mberacochea:
@hoelzer @fischer-hub this is amazing, thank you so much. This has been in our backlog for ages. I will go over the PR as soon as I can, but from a quick read it looks great.

virify.nf
```diff
@@ -477,11 +499,17 @@ workflow annotate {
...
-        taxonomy_annotations = assign.out.map { _, __, taxonomy -> taxonomy }.collect()
-        checkv_results = checkV.out.map { _, __, quality_summary, ___ -> quality_summary }.collect()
+        def i = 1 
+        def j = 1 
+        def k = 1 
+
+        // enumerate collected files to prevent file collision
+        viphos_annotations = annotation.out.map { _, __, annotations -> tuple(annotations, i++) }.collect(){it -> it[0]} //{ annotations, count -> "$annotations".replace('.', '_' + count + '.') }
+        taxonomy_annotations = assign.out.map { _, __, taxonomy -> tuple(taxonomy, j++) }.collect(){it -> it[0]} //{ taxonomy, count -> "$taxonomy".replace('.', '_' + count + '.') }
+        checkv_results = checkV.out.map { _, __, quality_summary, ___ -> tuple(quality_summary, k++) }.collect(){it -> it[0]} //{ quality_summary, count -> "$quality_summary".replace('.', '_' + count + '.') }
```
mberacochea, line 509: This is to support the -list of assemblies, correct?

virify.nf
```diff
@@ -477,11 +499,17 @@ workflow annotate {
...
 
         write_gff(
+          // when using --list with multiple samples and the first assembly is bad by chance this generates one bad GFF file right?
```
mberacochea, line 512: This is indeed a bug

mberacochea (changes requested):
Thanks @fischer-hub, great stuff. I left a few comments, nothing major.
The one thing I would like to add is a few more unit tests for parse_viral_pred.py, can you send me an example of the virsorter2 outputs?

fischer-hub:
> Thanks @fischer-hub, great stuff. I left a few comments, nothing major. The one thing I would like to add is a few more unit tests for parse_viral_pred.py, can you send me an example of the virsorter2 outputs?

Sure, will go over the changes asap  and attach some example output here!

virify.nf
fischer-hub: When running on a list of samples this was my first fix to prevent the `write_gff` file collision, but I think [this](https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/128/files#diff-271fae29b461eeb8b8010c1b16ce3d1daf0eb82f7e7d98d63372a0fc9f6f3e6cR9) should be sufficient as well since then the input files are enumerated and have different names too. (I just forgot to remove this afterwards, will check again)

mberacochea: I see. The change on the contigs.first() it a but more invovled as we need to .join() the assemblies with the corresponding annotations. Otherwise, the pipeline could mix results from different assemblies.

fischer-hub: Yes, true. Maybe its better then to keep the fix and handle the bug in a separate PR.

KateSakharova:
Fixed in #135

PR #129, opened 2024-07-20
Inver the order of the taxonomic lineages on the output files

### LINKED ISSUE #113: Taxonomic ranks are inverted
Hello there!

I've been testing the VIRify v2.0 and I realised that the taxonomic annotation on the GFF file has the ranks inverted. 

For instance:
taxonomy=Entomopoxvirinae;Poxviridae;Chitovirales

Should be:
taxonomy=Chitovirales;Poxviridae;Entomopoxvirinae

And ideally, it would be great to have the whole lineage like:
taxonomy=Viruses;Bamfordvirae;Nucleocytoviricota;Pokkesviricetes;Chitovirales;Poxviridae;Entomopoxvirinae

There are also some problems with names like Caudovirales which is shown in the NCBI taxonomy database as Caudoviricetes.

Thanks in advance!

Ales.

mberacochea:
This will close #113


hoelzer:
I think looks good... and then we would report the full taxonomy ranks which is even nicer. And in more meaningful order.

PR #130, opened 2024-08-23
Update README.md

cerebis:
Revised example commands to use $HOME rather /home/$USER, as the latter is not consistently valid across systems.


mberacochea:
Thanks @cerebis!

PR #134, opened 2024-10-01
Re-structure pipeline

KateSakharova:
Restructured pipeline to nf-core view
- created main.nf main entry
- added virify.nf to workflows and simplified
- all subworkflows wrapped into separate files
- modules moved to local/name/main.nf
- publishDir moved to modules.config
- container specified in each module main.nf 
- moved help to separate module
- moved warning to separate module
- removed CWL (woohoo)
- added samplesheet support
- removed --list option (replaced with samplesheet)
- replaced name with meta (resolved issues with --list option)
- added some nf-core tools

Now virify supports samplesheet (fq1 and fq2 are optional) 
id,assembly,fq1,fq2

For assembly as input:
If you have assembly as input you can use old argument --fasta - that will run pipieline for 1 assembly.
If you have more than 1 assembly - use samplesheet.

For raw_reads as input:
You need to use ONLY samplesheet (even for 1 run). I replaced --illumina option with --assemble (boolean). User should specify that flag to force assembly process (that flag will help user to understand what he is doing). _Hopefully, nobody uses it because it is experimental feature._


subworkflows/local/annotate.nf
```diff
@@ -93,25 +93,36 @@ workflow ANNOTATE {
...
     CHECKV(
-      predicted_contigs.combine( contigs.map { name, fasta -> fasta }),
+      predicted_contigs.join(contigs.map { meta, fasta -> fasta }),
       checkv_db
     )
 
-    viphos_annotations = ANNOTATION.out.map { _, __, annotations -> annotations }.collect()
-    taxonomy_annotations = ASSIGN.out.map { _, __, taxonomy -> taxonomy }.collect()
-    checkv_results = CHECKV.out.map { _, __, quality_summary, ___ -> quality_summary }.collect()
+    viphos_annotations = ANNOTATION.out.annotations.groupTuple().map{
```
mberacochea, line 100: Is this `groupTuple` required becase `ANNOTATION` return a list of files? I'm not sure I follow

mberacochea:
So far.. it look goods to me :) Thank you @KateSakharova

subworkflows/local/annotate.nf
KateSakharova: it is in development stage yet, I'm not sure what it returns either :D

mberacochea: Perfect hehe

PR #137, opened 2025-01-15
Feature/support protein predictions

KateSakharova:
Added support of using already predicted proteins as pipeline input.
- added script to split proteins.faa into HC, LC and PP files. Phophages processing is quite tricky because they have particular coords, not whole contig.
- fixed gff generator not to change coords for already predicted proteins **for prophages** (because they were initially predicted in contig coords)


assets/schema_input.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://json-schema.org/draft/2020-12/schema",
-    "$id": "https://raw.githubusercontent.com/ebi-metagenomics/miassembler/master/assets/schema_input.json",
+    "$id": "https://raw.githubusercontent.com/ebi-metagenomics/emg-viral-pipeline/dev/assets/schema_input.json",
```
mberacochea, line 3: We need to remember to update this `dev` -> `master`

KateSakharova: Yes, absolutely (bit now dev if quite different from master)

bin/split_proteins_by_categories.py
```diff
@@ -73,11 +74,18 @@ def check_coordinates(self, prodigal_info, prophage_info):
...
+        # Calculate overlap
+        overlap_start = max(start_protein, start_prophage)
+        overlap_end = min(finish_prophage, finish_protein)
+        intersection = max(0, overlap_end - overlap_start + 1)
+        # Make decision about protein
+        if intersection:
+            prophage_cov = intersection / prophage_length
+            protein_cov = intersection / protein_length
+            
+            if prophage_cov > 0.9 or protein_cov > 0.9:
```
Ales-ibt, line 87: Protein length will always be shorter that the prophage length. Evaluating the protein coverage should be enough. In the MAP we evaluate the MGE length because we can have small MGEs as short as 1 Kb, but in the case of prophages, that shouldn't be the case.

Ales-ibt (approved):
Thank you for this. I will update the protein coverage length on the next version of the MAP (wip) to be on the same page.

PR #138, opened 2025-02-19
Add annotations for momofy

KateSakharova:
Added `viral genes` and `kmer frequency` to checkV results attributes. Also added `virify_quality` {HC, LC, PP, unknown} for all annotations right after ID. _If quality is unknown - it seems an error in parsing. Better to report it to me to fix_

example:
```
NODE_5186_length_8751_cov_4.185028   VIRify viral_sequence 1    8751  .    .    .    ID=NODE_5186_length_8751_cov_4.185028|viral_sequence;virify_quality=LC;gbkey=mobile_element;mobile_element_type=phage_linear;checkv_provirus=No;checkv_quality=Low-quality;miuvig_quality=Genome-fragment;kmer_freq=1;viral_genes=15;taxonomy=Fletchervirus%3BEucampyvirinae
NODE_5194_length_8743_cov_5.186004   VIRify viral_sequence 1    8743
```


bin/write_viral_gff.py
```diff
@@ -156,11 +168,15 @@ def write_gff(
...
+                kmer_freq = row["kmer_freq"]
+                viral_genes = row["viral_genes"]
                 checkv_info = ";".join(
                     [
                         f"checkv_provirus={checkv_type}",
                         f"checkv_quality={checkv_quality}",
                         f"miuvig_quality={miuvig_quality}",
+                        f"kmer_freq={kmer_freq}",
+                        f"viral_genes={viral_genes}",
                     ]
```
Ales-ibt, line 180: Maybe add the 'checkv' prefix to `kmer_freq` and `viral_genes` attribute keys to denote where do they come from.

bin/write_viral_gff.py
```diff
@@ -233,6 +249,7 @@ def empty_if_number(string):
         # coordinates and attributes
         for contig_name, viral_sequence_types in viral_sequences.items():
             clean_contig_name = Record.remove_prophage_from_contig(contig_name)
+            quality = virify_quality[clean_contig_name] if clean_contig_name in virify_quality else "unknown"
```
Ales-ibt, line 252: In which case you could have "unknown" quality?

bin/write_viral_gff.py
KateSakharova: easy, yes

bin/write_viral_gff.py
KateSakharova: there shoud be no cases! That is a "check" for dummy mistake in code

PR #139, opened 2025-03-20
Prep for the release of version 3.0.0 (or 2.1.0)

mberacochea:
This is a mssive PR, @KateSakharova, @Ales-ibt did all the work here :); heavly inspired by the early work to support VirSorter2 by @fischer-hub

It includes:

- Adjust the code structure to a more nf-core like organization
- Using samplesheets - simplifies running the pipeline in bulk
- Adding VirSorter2 (optinally).
- Tweaks here and there


timrozday-mgnify (changes requested):
Looks good! I like the use of chunking for virsorter2 and the generation of sankey plots at the end.

I have one concern:
- For publishing `virsorter2` files in modules.config: what file extension does `virsorter2` use for the FASTA files, the tests have `.fa` which would not be published under the current config

Otherwise I just have small comments:
- `ASSEMBLE_ILLUMINA` looks non-functioning (correct me if I'm wrong). Consider removing in master.
- The `kaiju` module appears to be unused and should be removed (it was previously used in a filtering step that's been removed?). Remove modules.config entry as well.
- The `ch_multiqc_files` channel in `virify.nf` isn't used. Should either use it or lose it.

timrozday-mgnify:
Do you have the results of a successful pipeline run? Would be useful to check for things missing/strange

timrozday-mgnify:
Additionally PPR-Meta seems to be designed to handle raw metagenomic reads and so will probably work better run in this way rather than on assembled contigs. Given that the pipeline can accept un-assembled reads, it should be an option to run these straight into PPR-Meta (after QC etc...).

hoelzer:
> Additionally PPR-Meta seems to be designed to handle raw metagenomic reads and so will probably work better run in this way rather than on assembled contigs. Given that the pipeline can accept un-assembled reads, it should be an option to run these straight into PPR-Meta (after QC etc...).

Hey, I would not use PPR-Meta on the reads: 

> PPR-Meta was trained and tested using artificial contigs. We downloaded the accession list of prokaryote chromosomes, prokaryote plasmids, and phages from the NCBI genome database, and the corresponding genomes were downloaded according to the list. 

https://academic.oup.com/gigascience/article/8/6/giz066/5521157?login=true#494015635

Also, the initial step of VIRify is always assembly to get contigs (which are then classified as viruses or not by different tools, including PPR-Meta). It would be a way to first filter input reads with PPR-Meta and then assemble when the input is raw reads. But this could also lead to losing information for viruses that VS2 or VF could identify on a contig level. 

Of course what could be tested and discussed is replacing PPR-Meta eventually with another ML prediction tool - but that's probably another story for another PR : )

timrozday-mgnify:
> [https://academic.oup.com/gigascience/article/8/6/giz066/5521157?login=true#494015635](https://academic.oup.com/gigascience/article/8/6/giz066/5521157?login=true#494015635)

Thanks @hoelzer, I see now that the false positive rate is much higher from sequences in the 100-400bp range (figure 2). Another tool, another PR

modules/local/pprmeta/main.nf
Ales-ibt, line 1: This is not working. I am getting the following error in slurm:

```bash
ERROR ~ Error executing process > 'VIRIFY:DETECT:PPRMETA (ERZ13633678)'

Caused by:
  Process `VIRIFY:DETECT:PPRMETA (ERZ13633678)` terminated with an error exit status (1)


Command executed:

  export MCR_CACHE_ROOT="null/mcr_cache_root"
  [ -d "pprmeta" ] && cp pprmeta/* .
  ./PPR_Meta ERZ13633678_renamed_filt1500bp.fasta ERZ13633678_pprmeta.csv

Command exit status:
  1

Command output:
  (empty)

Command error:
  Fatal Internal Error: Unable to create preferences folder in null/mcr_cache_root/.mcrCache9.4/PPR_Me0/.matlab. Preferences folder location must be a valid full path. Ensure that your HOME environment variable is set to a full path.
```

Ales-ibt: Removing the line ` export MCR_CACHE_ROOT="null/mcr_cache_root"` from main.nf solved the problem.

hoelzer: Hey, I can confirm the same error on SLURM. Removing the line in `pprmeta/main.nf`, however, did not resolve the problem for me:

```bash
ERROR ~ Error executing process > 'VIRIFY:DETECT:PPRMETA (kleiner_2015)'

Caused by:
  Process `VIRIFY:DETECT:PPRMETA (kleiner_2015)` terminated with an error exit status (255)


Command executed:

  #export MCR_CACHE_ROOT="null/mcr_cache_root"
  [ -d "pprmeta" ] && cp pprmeta/* .
  ./PPR_Meta kleiner_2015_renamed_filt1500bp.fasta kleiner_2015_pprmeta.csv

Command exit status:
  255

Command output:
  (empty)

Command error:
  Could not access the MATLAB Runtime component cache. Details: fl:filesystem:SystemError; component cache root:; componentname: PPR_Meta
```

hoelzer:
Thanks for all the massive changes and updates! 

However, I can not run the pipeline locally on my mac right now because too much RAM is requested, e.g.

```bash
nextflow run main.nf --fasta tests/test_data/kleiner_2015.fasta -profile local,docker --output 2025-03-24-test -resume
```

```bash
ERROR ~ Error executing process > 'VIRIFY:DETECT:VIRFINDER (kleiner_2015)'

Caused by:
  Process requirement exceeds available memory -- req: 36 GB; avail: 16 GB
```

The memory is preconfigured now in the `configs/base.config` and this can be quite high for users running on a laptop or small work station. I manually changed the hardcoded memory numbers in this configuration file to run the pipeline on my mac with only 16 GB RAM. 

However, I think it's best to implement a way for users with low memory to also run the pipeline. Maybe the `params.memory` variable from the `nextflow.config` could be used in the `base.config` when the user sets a certain profile?

modules/local/pprmeta/main.nf
mberacochea: 😠 sorry about. that, that was my change (which I forgot to test).

KateSakharova:
@timrozday-mgnify thank you for review! That is very useful.

Publishing `virsorter2` is a bug, I'm fixing that. Good catch!
`kaiju` was removed as well. 

About assembly - that step is functioning [here](https://github.com/EBI-Metagenomics/emg-viral-pipeline/blob/dev/workflows/virify.nf#L100) if you specify `--assemble` argument. `multiqc` is functioning also on assemble subwf because all other tools are not in published multiqc tools. We need to create custom multiqc outputs for main processes but that takes time and will be done in next release.

KateSakharova:
@hoelzer Hi Martin, should that pipeline run KAIJU and PHANOTATE tools? I don't see those executed in main branch...

timrozday-mgnify:
Thank you for the changes, looks good!

I asked about assembly because:
When calling `modules/nf-core/spades/main.nf` `SPADES` the input channel is not set up properly. Will be a simple fix, just a map operator.
Otherwise I get the following message: `Process `VIRIFY:ASSEMBLE_ILLUMINA:SPADES` declares 3 input channels but 1 were specified`

KateSakharova:
@timrozday-mgnify  ahh, I didn't have chance to test that properly haha, I will fix in new PR covering your comments (I added you to review). Thanks!

hoelzer:
> @hoelzer Hi Martin, should that pipeline run KAIJU and PHANOTATE tools? I don't see those executed in main branch...

@KateSakharova no these are processes from an earlier phase of the pipeline development. Can be removed from my pov. We added PHANOTATE as an alternative to prodigal for better virus ORF prediction but this would need more testing and benchmarking.

mberacochea:
> > @hoelzer Hi Martin, should that pipeline run KAIJU and PHANOTATE tools? I don't see those executed in main branch...
> 
> @KateSakharova no these are processes from an earlier phase of the pipeline development. Can be removed from my pov. We added PHANOTATE as an alternative to prodigal for better virus ORF prediction but this would need more testing and benchmarking.

If you get to test PHANOTATE testing/benchmarking, it would be nice to make VIRIfy more agnostic of the tools it uses so we can swap versons/tools faster.

timrozday-mgnify (approved):
Looks great!

PR #142, opened 2025-03-26
Fix/prep release

KateSakharova:
- Fixing configs
- fix bug with virsorter2 outputs
- rm kaiju
- rm phanotate
- compress pdf plots
- added --publish_all flag, default is false, means only `08-final` dir would be published
- added structure example to readme


chrisAta (approved):
looks good to me!

PR #146, opened 2025-04-23
Dev

KateSakharova:


Ales-ibt (approved):
I can see this commit changes the logic to remove the hmmextend variable. It looks OK to me.

mberacochea (approved):
Alright, pending some more results checking I think we should merge this one.

PR #150, opened 2025-05-20
A few fixes on v3.0.0

mberacochea:
I'm testing this but hopefully it will fix a few issues in the pre-release 3.0.0.
- PPRmeta cache issues - changed the Matlab cache to the current dir (maybe this should be in under a flag as it's intrastructure dependent)
- VirSorter2 issues with the container and docker - permissions issues. I've rebuilt the container - also removed the on-the-fly conda created by the virsorter2 snakemake internals
- Nanoozo containers are missing - I've replaced those with a biocontainers wget one


modules/local/virsorter2/main.nf
```diff
@@ -20,24 +21,20 @@ process VIRSORTER2 {
...
-  virsorter run \
-            --db-dir ${database} \
-            -w virsorter2 \
-            -i ${fasta} \
-            -j ${task.cpus} \
-            ${args} \
-            all
+  virsorter run \\
+    --db-dir ${database} \\
+    --use-conda-off \\
```
KateSakharova, line 32: That argument is sitting in modules.config already https://github.com/EBI-Metagenomics/emg-viral-pipeline/blob/master/configs/modules.config#L684. I think that should be there.
That also reminded me that we probably should add a correct condition and do not include if pipeline will be launched with conda. There is a note about it in modules.config and ofc I forgot about it

KateSakharova (changes requested):
Thank you! I didn't have chance to test downloads modules and left those behind.. But those are also important.
I added one comment about virsorter2 arguments.

modules/local/virsorter2/main.nf
mberacochea: I overlooked that one, it should be there.. that's fine. 
With conda .. I think we would just need to add https://github.com/EBI-Metagenomics/containers/blob/master/virsorter2/2.2.4/env.yaml as `conda env.yml` and we can keep the `use-conda-off`.

mberacochea: I'll keep the one in args.. and remove the on in the module.

mberacochea: Done :)

Issue #148, opened 2025-05-07
Failed to use v3.0.0 for virus annotation


ErnieHsieh:
Hi,

I was running VIRify v3.0.0 on our HPC to annotate my vOTUs.fasta, and it showed an error: 'Missing project main script: /users/3058727/.nextflow/assets/EBI-Metagenomics/emg-viral-pipeline/virify.nf'. 

My script is as follows:

`# loading module
module load apps/anaconda3/2024.10/bin
module load apps/singularity/3.10.0
. /opt/gridware/depots/54e7fb3c/el8/pkg/apps/anaconda3/2024.10/bin/etc/profile.d/conda.sh
export CONDA_PKGS_DIRS=/mnt/scratch2/users/$USER/conda/pkgs
export CONDA_ENVS_PATH=/mnt/scratch2/users/$USER/conda/envs
export PATH=/mnt/scratch2/users/$USER/conda/envs/nextflow/bin/:$PATH

conda activate /mnt/scratch2/users/$USER/conda/envs/nextflow

nextflow run EBI-Metagenomics/emg-viral-pipeline -r v3.0.0 \
         --fasta '/mnt/scratch2/users/3058727/r1987gfs/novogene_test/HQ_vOTUs.fasta' \
         --length 0.2 \
         --onlyannotate \
         --output /mnt/scratch2/users/3058727/r1987gfs/novogene_test/virify_out/ \
         --workdir /mnt/scratch2/users/3058727/r1987gfs/novogene_test/work \
         -profile slurm,singularity` 

I was just wondering if the latest version v3.0.0 has been available already? I also tried to run -r v2.0.1 but it also failed to use (e.g., CheckV: v0.8.1 cannot found etc.). Shall I also need to change my commend?

Many thanks in advance for looking into this issue.

Best,
EH

mberacochea:
Hi @ErnieHsieh,

We are prepping to release v3.0.0, I was testing it and that is why I created the pre-release (whenever you are happy with it we will make the "official" release).

Something went wrong and I've fix it, I've recreated the pre-release for v3.0.0. Can you give it another try please?

V2.0.1 should work, what is the error you are getting?

Cheers

ErnieHsieh:
Hi @mberacochea ,

Thanks for getting back to me promptly.

Firstly, I re-tried to run `(nextflow) bash-4.4$ nextflow pull EBI-Metagenomics/emg-viral-pipeline` on HPC and it showed:

```
[...]

PR #153, opened 2025-06-14
Fix imgvr blast filter - GFF virify_quality validation issue

mberacochea:
I've also:
- Upgraded the requirements-test.txt packages
- Replaced some SeqIO.parse with the filename to a file handle; I was getting "ResourceWarning: unclosed file <_io.TextIOWrapper name='.../final-viral-combined.fa' mode='rt' encoding='UTF-8'>
  for record in SeqIO.parse(final_combined_fa_file, "fasta"):". I don't have the time to look into biopython so I used a context manager to avoid leaving file handlers opened by accident.
- In the blast filter removed a task attemp snippet, IMO it wasn't needed.

Should fix - https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/148


bin/parse_viral_pred.py
```diff
@@ -335,51 +337,52 @@ def merge_annotations(pprmeta, finder, sorter, sorter2, assembly, vs_cutoff):
...
+                    lc_predictions_contigs.append(seq_record)
+            else:
                 # Pro
-            elif seq_record.id in sorter_prophages:
-                # a contig may have several prophages
-                # for prophages write the record as it holds the
-                # sliced fasta
-                for record in sorter_prophages.get(seq_record.id):
-                    prophage_predictions_contigs.append(record.get_seq_record())
+                # TODO: check this decision because for VirSorter2 sequence can be in 2 categories
```
mberacochea, line 365:
```suggestion
```

PR #156, opened 2025-07-11
Fix/spaces in contignames issue

KateSakharova:
- Made coordinates `int` instead of `float`
- removed `postprocess` subwf (it was only restore). Added "short" names to original fasta and category fasta.
- changed `restore` function to add prophage metadata after contig name (before space) to make it unique.
Example,
`ERZ12593051.63 NODE_63_length_59433_cov_64.962579` would be `ERZ12593051.63|prophage-6654:18314`
it was `ERZ12593051.63 NODE_63_length_59433_cov_64.962579|prophage-6654:18314`
**checkV** takes only contig name before space and complained for duplicates if one contig represented twice but with different propahge regions in metadata


mberacochea (changes requested):
Fasta headers.. the bane of bioinformatics

workflows/virify.nf
```diff
@@ -5,8 +5,10 @@ include { samplesheetToList            } from 'plugin/nf-schema'
 /************************** 
 * MODULES
 **************************/
-
-include { MULTIQC                      } from '../modules/nf-core/multiqc'
+include { RESTORE as RESTORE_FOR_ONLY_ANNOTATE } from '../modules/local/restore'
+include { RESTORE as RESTORE_CATEGORY_FASTA    } from '../modules/local/restore'
+include { RESTORE as RESTORE_FILTERED_FASTA    } from '../modules/local/restore'
```
Ales-ibt, line 10: Loads of restoring here :D

bin/rename_fasta.py
```diff
@@ -126,6 +128,18 @@ def main():
...
+        help="Name of column FROM what do renaming",
+        type=str,
+        choices=['temporary', 'original', 'short']
+    )
+    restore_parser.add_argument(
+        "--to-restore",
+        help="Name of column TO what do renaming",
+        type=str,
+        choices=['temporary', 'original', 'short']
+    )
```
Ales-ibt, line 142: Could you add a comment line or include help message examples of the expected string for each name type?

Ales-ibt (approved):
Three steps of renaming seems a bit too much, but if it fix the issue with checkv, that's fine. I just added a comment asking for some examples of the expected strings in each type of renaming.

PR #160, opened 2025-10-23
Add check for many-Ns

KateSakharova:
If CheckV tool doesn't find proteins it fails on hmmsearch step. I had a case of contig with many NNNN in sequence. I've added removal procedure for that case

### DIFF
--- a/bin/parse_viral_pred.py
+++ b/bin/parse_viral_pred.py
@@ -400,7 +400,7 @@ def merge_annotations(pprmeta, finder, sorter, sorter2, assembly, vs_cutoff, out
         prophage_predictions_contigs,
         sorter_hc,
         sorter_lc,
-        sorter_prophages
+        sorter_prophages,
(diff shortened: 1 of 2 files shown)


bin/parse_viral_pred.py
```diff
@@ -372,14 +387,17 @@ def merge_annotations(pprmeta, finder, sorter, sorter2, assembly, vs_cutoff):
                     )
                 elif seq_record.id in pprmeta_lc and seq_record.id in finder_lowestc:
                     lc_predictions_contigs.append(seq_record)
-
+                    
+    print(f"Number of skipped contigs {count_skipped_contigs}")
```
mberacochea: We should collect the names of the contigs that fail and report them, otherwise contigs may be discarded and users may not understand why. I would print the names to a log file (sanity_check_failed_contigs.txt or similar)

PR #166, opened 2026-02-26
Do not consider fraggenescan proteins in the annotation module

mberacochea:
This patch fixes a problem when the script viral_contigs_annotation.py had when trying to assign fragenescan proteins to VIPHOGS.


tests/viral_contigs_annotation_fixtures/with_fragenescan_proteins/mixed_ratio.tsv
```diff
@@ -0,0 +1,5 @@
+query	ViPhOG	Abs_Evalue_exp	Taxon
+contig1_1	VPHOG_0001	1e-10	Caudo
+contig2_gene1	VPHOG_0002	1e-15	Herelle
```
Ales-ibt: Would this gene with viphog annotation be ignored because we do not have the coordinates in the header of the `mixed_proteins.faa` file?

Ales-ibt (commented):
Hello, thank you for this. I think we can use this opportunity to make a more flexible fix, otherwise we are not able to input non prodigal proteins into virify and we need it for other pipelines. 

For instance, what would happen if the user provides a proteins files where any protein have `#` as delimiter? If I am understanding well, all the proteins would be excluded and there's going to be no viphog annotation at all, affecting the final result. We would need to add a check and run prodigal internally if no useful proteins are found, but with this we miss the point of the option to provide external proteins. 

Instead of ignoring proteins without the expected header format, we could adapt the code to get the proteins sequences and coordinates from a proteins fasta file and a matching GFF file.

KateSakharova (changes requested):
@Ales-ibt we use prokka in catalogues, that PR is not related to that issue. But I have told @mberacochea that we need that fix as well.

I completelly agree with Ale, I think we should align virify with MAP and take proteins in GFF format, instead of FASTA. That was my bad that I decided to use fasta that is very inconvenient...

KateSakharova (approved):

PR #167, opened 2026-02-27
Bugfix/write gff missing contigs when user provided proteins

mberacochea:
Fix KeyError for contigs missing from assembly in write_viral_gff
Contigs present in annotation TSVs but absent from the assembly FASTA
caused a KeyError in three places in write_gff()

This happens when the user provide proteins, as the proteins could be
for contigs that were filtered out by the length filter VIRIfy use.
I've added a bit of code to make it fail if all the contigs are
discarded as this would a problem.

Adds a regression test with some file fixtures.


mberacochea:
I've accidenatly created this one on top of https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/166

chrisAta (approved):
looks good to me!

bin/write_viral_gff.py
```diff
@@ -307,13 +307,33 @@ def empty_if_number(string):
...
             sequence_regions.append((clean_contig_name, contig_length))
 
+    if missed_contigs > 0:
+        logging.warning(
+            f"{missed_contigs} contigs were not found in the assembly and were skipped"
+        )
+
+    if not sequence_regions:
+        raise ValueError(
+            "All the contigs that came from the annotated viral sequences were discarded."
```
Ales-ibt, line 336: But there's no way that this happen, right?  Because in the pipeline, contigs are first filtered by length end then the viral predictors are run

bin/write_viral_gff.py
```diff
@@ -580,4 +606,5 @@ def empty_if_number(string):
         virify_quality,
         contigs_len_dict,
         ena_mapping=ena_mapping,
+        user_proteins=args.use_proteins,
```
Ales-ibt: is the arg `use_proteins` or `user_proteins`?

Ales-ibt (approved):
Thanks for this. I left a couple of comments

bin/write_viral_gff.py
mberacochea: yes, this shouldn't happen. This is just defensive code in case there is something wrong elsewhere

bin/write_viral_gff.py
mberacochea: ha.. I'll rename this to be consistent.

PR #168, opened 2026-03-02
Feature/user proteins filter out proteins not in contigs

mberacochea:
Add protein-in-contigs filtering step and fix viral GFF generation for partial results

Introduce a new filter_proteins_in_contigs module that removes proteins
belonging to contigs excluded by length filtering AND the ones from non-prodigal origin.
This fixes a problem where user-provided proteins from length-filtered contigs were silently carried
forward into annotation, and those that came from FGS causing issues as they didn't match some of the
downstream steps.

Fixes:
- filter_proteins_in_contigs.py was referencing the wrong contigs file
- write_gff now uses only contigs with viral annotations (not the full
  filtered assembly), fixing mismatches in GFF output
- Add defensive checks in write_viral_gff.py for contigs absent from
  CheckV results, preventing silent failures on partial runs

Other changes:
- Add gzip support to filter_contigs_len.py
- Reformat Nextflow code to 4-space indentation (in most files but not all)


mberacochea:
I did a very dumb thing... I ran formatter and the number of spaces which makes reviwing the code harder. Apologies for that.

I'm running an assembly with this version and version 3.2 (no use proteins) to compare results.

bin/filter_contigs_len.py
```diff
@@ -16,25 +17,35 @@ def _filter_and_rename(iterator, threshold, run_id):
...
 
+def _fasta_opener(path: Path) -> tuple:
+    """Return the appropriate file opener and base stem for a FASTA file.
+
+    :param path: Path to a plain or gzipped FASTA file.
+    :return: Tuple of (opener callable, stem without compression/fasta suffixes).
+    """
+    if path.suffix == ".gz":
+        return gzip.open, Path(path.stem).stem
+    return open, path.stem
```
Ales-ibt, line 28: We don't use the hook anymore?
```python
        with fileinput.hook_compressed(path, "r", encoding="utf-8", errors="ignore") as input_table:
```

subworkflows/local/protein_prediction.nf
```diff
@@ -1,30 +1,29 @@
...
-include { PRODIGAL                    } from '../../modules/local/prodigal'
-
+include { PRODIGAL } from '../../modules/local/prodigal'
 
 workflow PREDICT_PROTEINS {
-
     take:
-    input_fastas
-    
+    input_fastas // (meta, set_name, fasta) or (meta, set_name, fasta, faa) when use_proteins
```
Ales-ibt, line 5: Thank you for the comments, this is very useful

modules/local/checkv/main.nf
```diff
@@ -1,28 +1,30 @@
...
-    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-        'https://depot.galaxyproject.org/singularity/checkv:1.0.1--pyhdfd78af_0':
-        'biocontainers/checkv:1.0.1--pyhdfd78af_0' }"
-        
+
+    tag "${meta.id} ${confidence_set_name}"
+
+    container "${workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container
+        ? 'https://depot.galaxyproject.org/singularity/checkv:1.0.1--pyhdfd78af_0'
+        : 'biocontainers/checkv:1.0.1--pyhdfd78af_0'}"
```
Ales-ibt, line 9: The biocontainer for v1.0.3 is available

Ales-ibt (approved):
It looks Ok to me. I just left a couple of comments.

bin/filter_contigs_len.py
mberacochea: yeah, I should.

subworkflows/local/protein_prediction.nf
mberacochea: @KateSakharova added loads of them

modules/local/checkv/main.nf
mberacochea: yeah, but I rather not change it on this PR

KateSakharova (approved):
Thank you for those fixes! 
It looks OK from glance. Please, just make sure new execution of pipeline works for HC, LC and PP examples (checking not only GFF but also krona/sankey plots). 

It also feels like it should be easy to add prokka support just adding one more regex like you had for prodigal `_PRODIGAL_PATTERN = re.compile(r"#\s*\d+\s*#\s*\d+\s*#\s*[+-]?1\s*#")`... but there is no need to do it right now

mberacochea:
I've compiled a comparison in our team Confluence space. Attached here as reference

[VIRIfy v3.2 - with third party proteins-040326-155641.pdf](https://github.com/user-attachments/files/25745143/VIRIfy.v3.2.-.with.third.party.proteins-040326-155641.pdf)

PR #169, opened 2026-04-08
nf-schema wasn't installed

mberacochea:
And the schema had some issues too.

I have to this this first.. but if someone spots any issues please flag them.

I'm doing this because I encounter a weird issues in the automation system... which I think is linked to the nextflow_schema.json issues I address in this PR.

### DIFF
--- a/nextflow_schema.json
+++ b/nextflow_schema.json
@@ -34,6 +34,23 @@
                     "help_text": "You will need to create a design file with information about the assembly/raw_reads running the pipeline. Use this parameter to specify its location. It has to be a comma-separated file with a minimum of 2 columns for assembly (ID, fasta) or 3 columns for raw reads (ID, fastq1, fastq2), and a header row.",
                     "fa_icon": "fas fa-file-csv"
                 }
+            },
+            "oneOf": [
+                {
+                    "required": [
+                        "fasta"
+                    ],
+                    "description": "An assembly file must be provided"
+                },
+                {
+                    "required": [
+                        "samplesheet"
+                    ],
+                    "description": "Assembly or raw reads should be provided in samplesheet"
+                }
+            ],
+            "errorMessage": {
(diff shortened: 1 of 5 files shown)


mberacochea:
The pipeline works.

nextflow_schema.json
```diff
@@ -34,19 +34,6 @@
                     "help_text": "You will need to create a design file with information about the assembly/raw_reads running the pipeline. Use this parameter to specify its location. It has to be a comma-separated file with a minimum of 2 columns for assembly (ID, fasta) or 3 columns for raw reads (ID, fastq1, fastq2), and a header row.",
                     "fa_icon": "fas fa-file-csv"
                 }
-            },
-            "oneOf": [
```
KateSakharova, line 38: It doesn't work? I feel like it should be described in schema as well, not only message in pipeline execution...

KateSakharova (approved):
Thank you!

nextflow_schema.json
mberacochea: I added this bit:
```
if (!params.samplesheet && !params.fasta) {
        error "You must specify either --samplesheet or --fasta as input."
}
```
it will fail with this error message

mberacochea: Sorry, I didn't answer your question. I didn't try with "oneOf", let me see if that works (between nf-validation and nf-schema I got a bit confused)

mberacochea: Indeed it works!:
```
Downloading plugin nf-schema@2.7.1
Launching `main.nf` [big_magritte] DSL2 - revision: 734ee1d06a

ERROR ~ Validation of pipeline parameters failed!

 -- Check '.nextflow.log' file for details
The following invalid input values have been detected:

* Value matches against more than one schema. Matched schema indexes [0, 1] ({"oneOf":"You must specify either an assembly file or samplesheet input."})
```

PR #171, opened 2026-04-17
Performance improvements of SPLIT_PROTEINS module

ochkalova:
Rewrite script for SPLIT_PROTEINS module to remove nested for loop. 
Also some refactoring done to improve readability and clarity
Running time went down from 43 hours on big assembly from ASG dataset to 5 mins. 

### Purpose of this script

This script takes:
- a FASTA file with selected subset contigs, and
- a FASTA file with all predicted proteins,

and creates a new proteins FASTA that contains only proteins that belong to those selected contigs.

#### How it works:

1. Read contigs and all proteins from FASTA files.
2. Build a lookup table from contig id to its protein records.
3. For each selected contig:
- find proteins that match its contig id,
- if contig is marked as prophage, keep only proteins that overlap the prophage region (or keep all for circular phages),
- update headers for proteins from contigs marked as prophage so they match prophage contig naming.
4. Write matched proteins to the output FASTA.
5. Validate results:
- fail if no proteins were written,
- fail if the same protein would be written more than once.

### DIFF
--- a/bin/split_proteins_by_categories.py
+++ b/bin/split_proteins_by_categories.py
@@ -20,7 +20,10 @@ import re
 import logging
 from collections import defaultdict
 from copy import deepcopy
+from typing import Dict, Iterable, List, Optional, Tuple
+
 from Bio import SeqIO
+from Bio.SeqRecord import SeqRecord
 
 
 def parse_args(argv):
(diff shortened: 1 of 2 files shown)


mberacochea (commented):
I will try to do a more thorough review, but from a quick read it looks sensible. If Kate reviews this.. and she is happy with it - merge away.

Two things:
- This version requires more memory, so make sure to adjust this in the process
- Please make sure this test works - https://github.com/EBI-Metagenomics/emg-viral-pipeline/blob/master/tests/test_split_proteins_by_categories.yml (I don't think it is being executed on the actions now)

bin/split_proteins_by_categories.py
```diff
@@ -91,49 +111,130 @@ def check_coordinates(self, prodigal_info, prophage_info):
...
+                    
+                    if protein_id in already_added_protein_ids:
+                        raise ValueError(f"Protein added more than once: {protein_id}")
+                    
+                    SeqIO.write(record, out_file, "fasta")
+                    already_added_protein_ids.add(protein_id)
+                    written_records += 1
+
+        if written_records == 0:
+            raise ValueError(
```
KateSakharova: Would it make the whole pipeline fail? I guess we do not want it.. @mbc? generating empty split_fasta.faa seems fine for production but ofc it might be just dangerous not to understand that issue can be in script itself and not in missing proteins. But unfortunately, for production we should generate empty fasta I guess... 🤔

KateSakharova (approved):
It looks good! Thank you, Sonya. The only one question I got about "no found proteins" exit code

bin/split_proteins_by_categories.py
ochkalova: 🤔 
Good point. I can imagine there will be only contigs without gene predictions in the input fasta. Would be better to show warning instead of failing completely.

KateSakharova: yes, I think warning should be better, but again we should be careful in production somehow (not related to that script :)

PR #172, opened 2026-05-05
Add proteins gff as pipeline input

KateSakharova:
That is turned into very massive PR, sorry! But the majority of fixes are not interesting.

**Important changes:**
- Add proteins.GFF as input
- split_proteins now returns FAA and GFF corresponding to chosen category (HC, LC, PP)
- Removed `|prophage` addition for protein sequences (contigs will have `|prophage` addition). It was used in final GFF generation process to adjust coords. Now we do not need that messy prefix in proteins because they can be easily matched with corresponding contig using GFF (tests for `split_proteins` reflect changes)
- removed viral regions from final GFF if checkv detected 0 viral genes
- I had to add `rename_prodigal` for mode `--use-proteins=false`. Prodigal replaces proteinID in GFF into 1_1, 1_2 but still reports contigID_1, cintigID_2 in FAA. That script brigs GFF IDs into FAA IDs space
- rewrote download DBs subworkflow

**Other changes:**
- Deleted filter_proteins_in_contigs
- fixed write_gff to support names of proteins
- replace FAA with GFF in `viral_contigs_annotation.py` (to grep coordinates from GFF instead of header). Test files were changed from FAA to GFF
- removed/changed env variables for 2 processes
- changed publishDir in downloadDBs processes
- simplified modules.config
- fix ratio_evalue step to return empty file and do not break futher processing for category of predictions

**Rewrote write_viral_gff.py** script without robots. Claude helped me to summarise changes

Bug fixes
                                                                                                                                                                                
  1. CDS quality always "unknown" — virify_quality is keyed by the full contig ID (with |prophage- suffix), but the CDS loop was looking up by clean name. Fixed by storing the
  original contig ID as a 6th element in each cds_annotations entry and using virify_quality.get(original_contig, "unknown").                                                   
  2. All proteins missing from GFF — cds_annotations.append(...) sat inside if best_hit != "No hit":, so proteins with no ViPhOG hit were never written. Issue https://github.com/EBI-Metagenomics/emg-viral-pipeline/issues/92. Moved the append     
  unconditional; viphog_annotation stays empty string for no-hit proteins.                                                                                                      
  3. Sanity check filter removed — A block that cross-filtered virify/checkv/taxonomy files to only keep confidence levels present in all three was deleted. All provided files
  are now processed as-is.                                                                                                                                                      
  4. Six bugs found in the refactored version:                                                                                                                                
    - define_virify_quality — setdefault was outside the row loop (wrong indentation), so only the last contig per file was saved                                               
    - main() unpacked 3 values from get_annotation_results but it returned 2 → ValueError                                                                                       
    - define_virify_quality called with checkv_files instead of virify_files                                                                                                    
    - get_sequence_regions referenced contigs_len_dict as a global — not a parameter → NameError                                                                                
    - get_proteins_from_gff called with bare gff_files instead of args.gff_files → NameError                                                                                    
    - Debug print(contig_name, viral_sequence_types) left inside write_gff                                                                                                                                                                                                                                                                       

**New functionality**                                                                                                                                                             
                                                                                                                                                                              
  5. --gff / -g argument added — accepts per-category GFF3 files from split_proteins or prodigal.
  6. read_category_gff_contigs — reads all seqids from those GFF files.
  7. get_proteins_from_gff — supplements viral_sequences with any contigs that appear in the per-category GFFs but have no proteins (and therefore no annotation TSV rows).


Ales-ibt:
One thing regarding the 1-based vs 0-based coordinate system issue. The standard for GFF is  1-based fully closed start-end, meaning that features starting on 0 or ending on a position outside the contig length will brake the validation step

assets/schema_input.json
```diff
@@ -29,14 +29,22 @@
...
+                "errorMessage": "The file with proteins detected for the assembly needs to be a gff3 file (optionally compressed with .gz)"
+            },
+            "proteins_faa": {
+                "type": "string",
+                "pattern": "^\\S+\\.(faa|fasta)(\\.gz)?$",
+                "errorMessage": "The file with proteins detected for the assembly in fasta format (optionally compressed with .gz)"
             }
         },
         "required": ["id"],
+        "dependentRequired": {
```
Ales-ibt, line 44: Very useful, I will use it in the MAP :D

bin/split_proteins_by_categories.py
```diff
@@ -103,30 +104,15 @@ def check_coordinates(self, prodigal_info: str, prophage_info: str) -> bool:
             protein_cov = intersection / protein_length
             
             if prophage_cov > 0.9 or protein_cov > 0.9:
-                self.logger.debug(f"Protein {prodigal_info[4:20]} more than 90% inside {prophage_info}")
+                self.logger.debug(f"Protein {protein_id} more than 90% inside {prophage_info}")
                 return True
             else:
-                self.logger.debug(f"Protein {prodigal_info[4:20]} intersects {prophage_info}")
+                self.logger.debug(f"Protein {protein_id} intersects {prophage_info}")
```
Ales-ibt: Probably something more informative "Protein partially intersect, not part of the prophage: {protein_cov}%"

bin/split_proteins_by_categories.py
```diff
@@ -191,45 +219,57 @@ def grep_proteins(self) -> None:
...
                 for protein_record in matching_proteins:
                     protein_id = protein_record.id
 
-                    # TODO: If we want to simulate proper Prodigal naming scheme we need to change proteinID
-                    # to new numbers starting with 1. Otherwise we have some skipped numbers in the output fasta
-                    if prophage_addition:
-                        self.logger.debug(f"Checking coordinates for protein {protein_id} against prophage info {prophage_addition}")
-                        protein_info, protein_number = self._parse_protein_id(protein_record.description, contig_name)
-                        if not self.check_coordinates(protein_info, prophage_addition):
+                    if prophage_addition is not None:
```
Ales-ibt, line 248: Why `if is not None` instead of the `if prophage_addition:` as it was before?

bin/viral_contigs_annotation.py
```diff
@@ -10,40 +10,68 @@
...
-        # Skip proteins that don't have Prodigal format (e.g., frageneScan proteins)
-        if not protein_prop:
-            continue
+        if gff_data is not None and protein.id in gff_data:
+            info = gff_data[protein.id]
+            contig_id = info['contig']
+            protein_prop = [protein.id, info['start'], info['end'], info['strand']]
+            query_id = protein.id
+        else:
+            # Fallback: extract coordinates from Prodigal-format description
```
Ales-ibt: So here you are catching the protein ID from the string in prodigal proteins fasta file. I think this could be a process by it's own to standardise the headers of the proteins fasta file when necessary no matter if they comes from the user or are internally generated. Just an idea to make it more transparent.

modules/local/get_db/virfinder.nf
```diff
@@ -2,12 +2,7 @@ process virfinderGetDB {
   label 'process_low'    
   container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
   
-  if (params.cloudProcess) {
```
Ales-ibt, line 5: Out of curiosity, why are you removing cloud profile?

subworkflows/local/protein_prediction.nf
```diff
@@ -2,28 +2,30 @@ include { PRODIGAL } from '../../modules/local/prodigal'
...
     category_fasta = channel.empty()
 
     if (params.use_proteins) {
-        category_fasta = input_fastas.map { meta, type, fasta, _faa -> tuple(meta, type, fasta) }
-
-        // skip prodigal step and use existing faa-s
-        proteins = input_fastas.map { meta, type, _fasta, faa -> tuple(meta, type, faa) }
+        category_fasta = input_fastas.map { meta, type, fasta, _faa, _gff -> tuple(meta, type, fasta) }
+        proteins       = input_fastas.map { meta, type, _fasta, faa, _gff -> tuple(meta, type, faa) }
+        proteins_gff   = input_fastas.map { meta, type, _fasta, _faa, gff -> tuple(meta, type, gff) }
```
Ales-ibt, line 17: Or a multimap :D

Ales-ibt (commented):
Nice job! Thanks for this. I know it is WIP, so I will just left the comments.

modules/local/get_db/virfinder.nf
KateSakharova: it is not removed, just reworded
`publishDir "${params.databases}/virfinder", pattern: "VF.modEPV_k8.rda", mode: params.cloudProcess ? 'copy' : 'symlink'`

bin/rename_prodigal.py
```diff
@@ -0,0 +1,118 @@
...
+    with open(proteins, "r") as file_in:
+        for line in file_in:
+            if not line.startswith(">"):
+                continue
+            protein_id = line.strip().split(" ")[0].replace(">", "")
+            protein_names.add(protein_id)
+    return protein_names
+
+
+def rename_proteins_in_gff(gff: str, protein_names: set[str], output: str) -> None:
```
mberacochea, line 67: Can you please add a doc string?

bin/rename_prodigal.py
```diff
@@ -0,0 +1,118 @@
...
+    with open(gff, "r") as file_in, open(output, "w") as file_out:
+        for line in file_in:
+            if line.startswith("#"):
+                file_out.write(line)
+                continue
+            fields = line.strip().split("\t")
+            if len(fields) != 9:
```
mberacochea, line 87: We should either fail if the number fields is different than 9... it really shouldn't happen

bin/rename_prodigal.py
```diff
@@ -0,0 +1,118 @@
...
+            if fields[2] != "CDS":
+                file_out.write("\t".join(fields) + "\n")
+                continue
+            attrs, _ = parse_attrs(fields[8])
+            protein_id = attrs.get("ID")
+            if not protein_id:
+                logging.warning("No ID in: %s", "\t".join(fields))
+                file_out.write("\t".join(fields) + "\n")
+                continue
+            if not re.fullmatch(r"\d+_\d+", protein_id):
```
mberacochea: A comment here explaning why this "arbitraty" pattern. (d_d)

bin/rename_prodigal.py
```diff
@@ -0,0 +1,118 @@
...
+                # ID is already in long format or unrecognised pattern; pass through
+                file_out.write("\t".join(fields) + "\n")
+                continue
+            protein_suffix = protein_id.rsplit("_", 1)[-1]
+            new_protein = f"{contig_id}_{protein_suffix}"
+            if new_protein not in protein_names:
+                logging.warning("%s was not found in FAA", new_protein)
```
mberacochea, line 108: Shuld we fail here instead?

bin/viral_contigs_annotation.py
```diff
@@ -5,47 +5,68 @@
...
+            if line.startswith('#'):
+                continue
+            cols = line.strip().split('\t')
+            if len(cols) != 9 or cols[2] != 'CDS':
+                continue
+            attrs = {}
+            for part in cols[8].rstrip(';').split(';'):
+                if '=' in part:
+                    k, v = part.split('=', 1)
+                    attrs[k.strip()] = v.strip()
```
mberacochea: There is a bit of repetition here, maybe extract these 2 methods (and add some guards to make the code more defensive).

bin/viral_contigs_annotation.py
```diff
@@ -5,47 +5,68 @@
...
+            protein_id = attrs.get('ID', '').strip()
+            if protein_id:
+                seqid = cols[0]
+                if '|prophage-' in seqid:
+                    # Prodigal can use <contigNumber_proteinNumber> like 1_1 
+                    # and do not apply _proteinNumber to real contig name
+                    if re.fullmatch(r"\d+_\d+", protein_id):
+                        # if GFF ID has pattern <contigNumber_proteinNumber>
+                        protein_suffix = protein_id.rsplit('_', 1)[-1]
```
mberacochea: The regex should probably be a constant. And make the code more defensive... I worry about errors not being caught.

bin/write_viral_gff.py
```diff
@@ -28,251 +151,246 @@ def get_ena_contig_mapping(ena_contig_file):
...
-                checkv_info = ";".join(
-                    [
-                        f"checkv_provirus={checkv_type}",
-                        f"checkv_quality={checkv_quality}",
-                        f"checkv_miuvig_quality={miuvig_quality}",
-                        f"checkv_kmer_freq={kmer_freq}",
-                        f"checkv_viral_genes={viral_genes}",
-                    ]
-                )
+                if int(row['viral_genes']) == 0 and row['checkv_quality'] == "Not-determined":
```
mberacochea: Should we be a bit more defensive here:

```python
viral_genes_count = row['viral_genes'].strip()

if viral_genes_count == '':
    raise ValueError("viral_genes is empty")
```

bin/write_viral_gff.py
```diff
@@ -28,251 +151,246 @@ def get_ena_contig_mapping(ena_contig_file):
...
+        )
+
+    if contigs_without_checkv:
+        logging.warning(
+            f"{len(contigs_without_checkv)} viral contigs have no CheckV results "
+            "and will use placeholder NA values: "
+            + ", ".join(sorted(contigs_without_checkv))
+        )
+    
+    if not_determined:
```
mberacochea, line 363: Should we skip these?. There is potential for unkown things in here that are potentially interesting; maybe we should mark them as `checkv_quality=unkown` or something like that?

configs/modules.config
```diff
@@ -12,694 +12,413 @@ process {
...
-                    enabled: params.publish_all,
-                    mode: params.publish_dir_mode,
-                    failOnError: false,
-                    pattern: "*final-viral*.{tsv,fa}"
-                ]
+            [
+                path: { "${params.output}/${meta.id}/${params.virusdir}/virsorter2" },
+                enabled: params.publish_all,
+                mode: params.publish_dir_mode,
+                failOnError: false,
```
mberacochea: `failOnError: false,` this was already here, I missed it.
We risk having a pipeline that completes but doesn't copy all the files to the output directory.

configs/modules.config
```diff
@@ -12,694 +12,413 @@ process {
...
-                        return "${meta.id}/${params.plotdir}/sankey/${output_file.name}";
-                    }
-                },
+                path: { "${params.output}/${meta.id}/${params.plotdir}/sankey" },
                 enabled: params.publish_all,
                 mode: params.publish_dir_mode,
                 failOnError: false,
-                pattern: "*.sankey.filtered-${params.sankey}.json"
+                pattern: "*.sankey.filtered-${params.sankey}.json",
+                saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
```
mberacochea: I don't think you need the saveAs to ignore the versions.yml because you have pattern already (same in other publishDir directives)

modules/local/get_db/checkv.nf
```diff
@@ -2,12 +2,7 @@ process checkVGetDB {
     label 'process_low'    
     container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
     
-    if (params.cloudProcess) { 
-        publishDir "${params.databases}/checkv", mode: 'copy' 
-    }
-    else { 
-        storeDir "${params.databases}/checkv" 
-    }
+    publishDir "${params.databases}/checkv", mode: params.cloudProcess ? 'copy' : 'symlink'
```
mberacochea: For a future release we need to get rid of this "cloudProcess" thing; I find it confusing.

modules/local/parse/main.nf
```diff
@@ -30,12 +30,12 @@ process PARSE {
...
 
   input:
-  tuple val(meta), path(fasta), val(contig_number), path(virfinder), path(virsorter), path(pprmeta)
+  tuple val(meta), path(fasta), path(virfinder), path(virsorter), path(pprmeta)
 
   output:
   tuple val(meta), path("*.fna"), path('virsorter_metadata.tsv'), path("${meta.id}_virus_predictions.stats"), optional: true
 
-  when: contig_number.toInteger() > 0
+  when: fasta.size() > 0
```
mberacochea, line 38:
```suggestion
  when: fasta.countFasta() > 0
```
it feels more appropiate

modules/local/pprmeta/main.nf
```diff
@@ -4,14 +4,14 @@ process PPRMETA {
...
-  tuple val(meta), path(fasta), val(contig_number)
+  tuple val(meta), path(fasta)
   path pprmeta_git
 
   output:
   tuple val(meta), path("${meta.id}_pprmeta.csv")
 
   when:
-  contig_number.toInteger() > 0
+  fasta.size() > 0
```
mberacochea, line 14:
```suggestion
  fasta.countFasta() > 0
```

modules/local/virfinder/main.nf
```diff
@@ -7,18 +7,18 @@ process VIRFINDER {
...
+  tuple val(meta), path(fasta)
   path model
 
   output:
-  tuple val(meta), path("${meta.id}.txt")
+  tuple val(meta), path("${meta.id}.${fasta.baseName}.txt"), emit: result_tsv
 
   when:
-  contig_number.toInteger() > 0
+  fasta.size() > 0
```
mberacochea, line 17:
```suggestion
  fasta.countFasta() > 0
```

modules/local/virsorter/main.nf
```diff
@@ -42,14 +42,14 @@ process VIRSORTER {
...
-  tuple val(meta), path(fasta), val(contig_number)
+  tuple val(meta), path(fasta)
   path database
 
   output:
   tuple val(meta), path("*")
 
   when:
-  contig_number.toInteger() > 0
+  fasta.size() > 0
```
mberacochea, line 52:
```suggestion
  fasta.countFasta() > 0
```

modules/local/virsorter2/main.nf
```diff
@@ -7,15 +7,15 @@ process VIRSORTER2 {
...
+  tuple val(meta), file(fasta)
   path database
 
   output:
   tuple val(meta), path("*.final-viral-score.tsv"),    emit: score_tsv
   tuple val(meta), path("*.final-viral-boundary.tsv"), emit: boundary_tsv
   tuple val(meta), path("*.final-viral-combined.fa"),  emit: combined_fa
 
-  when: number_of_contigs.toInteger() > 0
+  when: fasta.size() > 0
```
mberacochea, line 18: This things, I proposed `countFasta` but it will be slower so `.size()` is faster. But, for a non-compressed version I think is fine... but what a compressed file won't be size 0

subworkflows/local/annotate.nf
```diff
@@ -54,6 +54,7 @@ workflow ANNOTATE {
 
     category_fasta = PREDICT_PROTEINS.out.category_fasta
     proteins       = PREDICT_PROTEINS.out.proteins
```
mberacochea: Maybe we should rename this to `proteins_fasta`

subworkflows/local/assemble_illumina.nf
```diff
@@ -28,7 +28,7 @@ workflow ASSEMBLE_ILLUMINA {
  
         // assembly
         SPADES(
```
mberacochea, line 30: Honestly.. I think we should drop this local flow. Our team never uses this, and really don't test it

subworkflows/local/detect.nf
```diff
@@ -11,34 +11,37 @@ include { CONCATENATE_FILES as CONCATENATE_FILES_SCORE               } from '../
...
+  renamed_assembly
   virsorter_db
   virfinder_db
   pprmeta_git
 
   main:
-
+  // chunk fasta by default by 500Mb
+  chunked_ch = renamed_assembly.flatMap { meta, fasta ->
+     def chunks = fasta.splitFasta(file: true, size: ${params.chunk_fasta_size})
```
mberacochea:
```suggestion
     def chunks = fasta.splitFasta(file: true, size: params.chunk_fasta_size)
```

subworkflows/local/detect.nf
```diff
@@ -67,12 +70,19 @@ workflow DETECT {
       }
   }
 
-  VIRFINDER(renamed_assembly_and_contigs_count, virfinder_db)
+  VIRFINDER(chunked_ch, virfinder_db)
+
+  CSVTK_CONCAT (
```
mberacochea:
```suggestion
  CSVTK_CONCAT(
```
_very important_

subworkflows/local/protein_prediction.nf
```diff
@@ -1,29 +1,40 @@
...
-    input_fastas // (meta, set_name, fasta) or (meta, set_name, fasta, faa) when use_proteins
+    input_fastas // (meta, set_name, fasta) or (meta, set_name, fasta, faa, gff) when use_proteins
 
     main:
 
     proteins       = channel.empty()
+    proteins_gff   = channel.empty()
     category_fasta = channel.empty()
 
     if (params.use_proteins) {
```
mberacochea, line 14: [not for this PR] but I think we should push this `use_proteins` to the samplesheet

subworkflows/local/protein_prediction.nf
```diff
@@ -1,29 +1,40 @@
...
     else {
         category_fasta = input_fastas
 
         // ORF detection --> prodigal
-        PRODIGAL(input_fastas)
-        proteins = PRODIGAL.out
+        PRODIGAL (
+            input_fastas
+        )
+        proteins     = PRODIGAL.out.proteins
```
mberacochea:
```suggestion
        proteins = PRODIGAL.out.proteins
```

subworkflows/local/protein_prediction.nf
```diff
@@ -1,29 +1,40 @@
...
+        proteins       = input_fastas.map { meta, type, _fasta, faa, _gff -> tuple(meta, type, faa) }
+        proteins_gff   = input_fastas.map { meta, type, _fasta, _faa, gff -> tuple(meta, type, gff) }
     }
     else {
         category_fasta = input_fastas
 
         // ORF detection --> prodigal
-        PRODIGAL(input_fastas)
-        proteins = PRODIGAL.out
+        PRODIGAL (
```
mberacochea:
```suggestion
        PRODIGAL(
```
We should migrate to pyrodigal (?)

tests/test_rename_prodigal.py
```diff
@@ -0,0 +1,161 @@
...
+
+        output_path = _write_tmp("")
+        try:
+            rename_proteins_in_gff(gff, protein_names, output_path)
+            with open(output_path) as f:
+                for line in f:
+                    if line.startswith("#") or not line.strip():
+                        continue
+                    fields = line.strip().split("\t")
+                    if len(fields) == 9 and fields[2] == "CDS":
```
mberacochea, line 105: We need to start using a gff parser of sorts, this code is repeated many times.

I think the toolkit is not miles away from being able to provide this. Anyway... just a side rant

tests/test_split_proteins_by_categories.py
```diff
@@ -0,0 +1,163 @@
...
+from pathlib import Path
+import hashlib
+import tempfile
+import unittest
+
+def md5sum(path):
+    return hashlib.md5(path.read_bytes()).hexdigest()
+
+
+class SplitProteins(unittest.TestCase):
```
mberacochea, line 19: I think it would be easier to test these with nf-test instead because then we can test the container and whatnot

modules/local/split_proteins/main.nf
```diff
@@ -3,26 +3,36 @@ process SPLIT_PROTEINS {
     label 'process_single'
     
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
-                'https://depot.galaxyproject.org/singularity/biopython:1.86' :
-                'quay.io/biocontainers/biopython:1.86' }"
+                'https://depot.galaxyproject.org/singularity/biopython:1.84' :
+                'quay.io/biocontainers/biopython:1.84' }"
```
mberacochea: Why the downgrade of biopython?

modules/local/rename_prodigal/main.nf
```diff
@@ -0,0 +1,19 @@
+process RENAME_PRODIGAL {
+    label 'process_single'
+    tag "${meta.id} ${confidence_set_name}"
+    container 'quay.io/microbiome-informatics/virify-python3:1.2'
```
mberacochea, line 4: Can't we use a stock python container?

workflows/virify.nf
```diff
@@ -160,17 +160,13 @@ workflow VIRIFY {
...
-        FILTER_PROTEINS_IN_CONTIGS(
-            faa.join(assembly_with_short_contignames)
-        )
-
-        SPLIT_PROTEINS(category_fasta.groupTuple().join(FILTER_PROTEINS_IN_CONTIGS.out).transpose())
+        SPLIT_PROTEINS(category_fasta.groupTuple().join(faa).transpose())
 
-        proteins_ch = SPLIT_PROTEINS.out
+        proteins_ch = SPLIT_PROTEINS.out.proteins
+            .join(SPLIT_PROTEINS.out.gff, by: [0, 1])
```
mberacochea: I think that instead of doing this join here, we can modify SPLIT_PROTEINS to return a tuple(meta, proteins_faa, proteins_gff) instead.

mberacochea (changes requested):
Massive work @KateSakharova :)

I will the python scripts another read.. so I may post more comments today.

modules/local/split_proteins/main.nf
ochkalova: Because there is no singularity container for `biopython:1.86` available :(

mberacochea: ahh.. Seqera containers to the rescue:
- docker: community.wave.seqera.io/library/pip_biopython:702fef869ad99456
- singularity: oras://community.wave.seqera.io/library/pip_biopython:746a711789280f4a

bin/split_proteins_by_categories.py
KateSakharova: I think it is more explicit because if `prophage_addition` would be false - then your condition will pass but mine not

subworkflows/local/assemble_illumina.nf
KateSakharova: Maybe.. I will not do it now because it requires many changes, leave it for next review

bin/write_viral_gff.py
KateSakharova: That is good point. I will create an Issue about it. I will not change that now because it changes functionality but that version was already used for catalogues generation

KateSakharova:
I have merged that monster to create more feature branches with fixes in the future

PR #176, opened 2026-06-09
Add chunking of input fasta for VirFinder and PPRmeta

ochkalova:
I transfered Kate's changes from #172 for chunking to include them to a small release that we can use in production (3.3.2). 

- Kate did chunking for VirFinder, I added one for PPRmeta
- I also had to redefine circularity to `False` in the `bin/parse_viral_pred.py` to fix this bug from VirSorter2 https://github.com/jiarong/VirSorter2/issues/251
- Update containers in SPLIT_PROTEINS because current singularity container is not accessible

### DIFF
--- a/bin/parse_viral_pred.py
+++ b/bin/parse_viral_pred.py
@@ -284,6 +284,11 @@ def parse_virus_sorter2(sorter_files, vs_cutoff):
             
             if "partial" in record.id:
                 record.id = clean_name
+                # In VirSorter2 records inherit "circular" label from the contig, which is incorrect for partial sequences 
+                # as they are extracted from the contig and might not be circular themselves.
+                # https://github.com/jiarong/VirSorter2/issues/251
+                # We rewrite "circular" to linear for partial sequences to avoid misinterpretation of the results.
+                circular = False
                 # add the prophage position within the contig
                 prophages.setdefault(clean_name, []).append(
                     Record(record, "prophage", circular, prange)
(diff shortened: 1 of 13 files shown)


bin/parse_viral_pred.py
```diff
@@ -284,6 +284,11 @@ def parse_virus_sorter2(sorter_files, vs_cutoff):
             
             if "partial" in record.id:
                 record.id = clean_name
+                # In VirSorter2 records inherit "circular" label from the contig, which is incorrect for partial sequences 
+                # as they are extracted from the contig and might not be circular themselves.
+                # https://github.com/jiarong/VirSorter2/issues/251
+                # We rewrite "circular" to linear for partial sequences to avoid misinterpretation of the results.
+                circular = False
```
Ales-ibt, line 291: Great! Thank you for raising the issue with the authors of virsorter2 as well. I will bring this change to the `feature/accept-different-proteins` branch as well

Ales-ibt (approved):
Thank you for this Sonya, it looks clean to me.

bin/parse_viral_pred.py
mberacochea: Good catch

mberacochea (approved):
Thanks @ochkalova 🥇 . Feel free to merge and release :)

KateSakharova (approved):
:shipit:

PR #177, opened 2026-06-24
Update database paths and a few configs

Ge94:
Went down a rabbit hole to change all paths generated in VIRify... I could have saved some of these by changing the configs alone, but since I was here 😅️ A summary:

- Moved `pprmetaGet` to `get_db`
- Removed all `databases` instances from path generators
- Adapted `download_databases` accordingly
- Behaviour tested with MGnify configs

### DIFF
--- a/modules/local/get_db/checkv.nf
+++ b/modules/local/get_db/checkv.nf
@@ -2,7 +2,7 @@ process checkVGetDB {
     label 'process_low'    
     container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
     
-    publishDir "${params.databases}/checkv", mode: params.cloudProcess ? 'copy' : 'symlink'
+    publishDir "${params.checkv}", mode: params.cloudProcess ? 'copy' : 'symlink'
       
     output:
         path("checkv-db-v*", type: 'dir')
--- a/modules/local/get_db/meta.nf
+++ b/modules/local/get_db/meta.nf
@@ -2,7 +2,7 @@ process metaGetDB {
     label 'process_low'    
     container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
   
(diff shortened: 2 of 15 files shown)


modules/local/get_db/checkv.nf
```diff
@@ -2,7 +2,7 @@ process checkVGetDB {
     label 'process_low'    
     container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
     
-    publishDir "${params.databases}/checkv", mode: params.cloudProcess ? 'copy' : 'symlink'
+    publishDir "${params.checkv}", mode: params.cloudProcess ? 'copy' : 'symlink'
```
KateSakharova, line 5: ahh it should be `checkv_db` or something saying it is a database (and in other placese as well) 🤦‍♀️  but it will require change in configs, maybe ignore for now

modules/local/get_db/meta.nf
```diff
@@ -2,7 +2,7 @@ process metaGetDB {
     label 'process_low'    
     container 'quay.io/biocontainers/gnu-wget:1.18--hb829ee6_10'
   
-    publishDir "${params.databases}/models", pattern: "additional_data_vpHMMs_${params.meta_version}.tsv", mode: params.cloudProcess ? 'copy' : 'symlink'
+    publishDir "${params.meta}", pattern: "additional_data_vpHMMs_${params.meta_version}.tsv", mode: params.cloudProcess ? 'copy' : 'symlink'
```
KateSakharova, line 5: what a terrible params name (params.meta) omg... it is from ancient times 🤦‍♀️

KateSakharova (approved):
Thank you so much, Germana! Added my 🤦‍♀️ -s

modules/local/get_db/checkv.nf
Ge94: :homer-disappearing.......

PR #180, opened 2026-07-15
fix rename missing output

KateSakharova:
- map.tsv and original.fasta were saved to workdir instead of output_dir


Ales-ibt (approved):
Thanks Kate, It makes sense, the pipeline was publishing resuts in the launching dir. This should fixed the issue.

PR #181, opened 2026-07-16
Remove assembling

KateSakharova:
- Removed assemble options: `--fasta` and `--assemble`. Virify will not support reads anymore and will accept only samplesheet as input.
- Fixed docs. Initially I wanted to just remove assembly part but figured out that I hate that README and restructured it to 4 files. I didn't re-write text, just moved to different files.
- I'm trying to reproduce schema in SVG format in order to make it editable for future fixes and add protein input there


docs/usage.md
```diff
@@ -0,0 +1,177 @@
...
+nextflow run EBI-Metagenomics/emg-viral-pipeline -r v3.0.0 \
+    --samplesheet samplesheet.csv \
+    --cores 4 -profile local,docker
+```
+
+__Please note__ that in particular the following parameters are important to handle where Nextflow writes files.
+
+* `--workdir` or `-w` (here your work directories with intermediate data will be saved)
+* `--databases` (here your databases will be saved and the workflow checks if they are already available under this path)
+* `--singularity_cachedir` (here Singularity containers will be cached, not needed for Docker, default path: `./singularity`)
```
mberacochea: (not now) but we need to get rid of this option

docs/usage.md
```diff
@@ -0,0 +1,177 @@
...
+* `--publish_all` - publish the expanded, per-step output folder structure (see [Output](output.md)) instead of just `08-final`.
+* `--output <dir>` - change the name/location of the results folder (default: `results`).
+
+## Profiles
+
+Nextflow uses a merged profile handling system so you have to define an executor (e.g., `local`, `lsf`, `slurm`) and an engine (`docker`, `singularity`) to run the pipeline according to your needs and infrastructure. 
+
+Per default, the workflow runs locally (e.g., on your laptop) with Docker. When you execute the workflow on a HPC you can for example switch to a specific job scheduler and Singularity instead of Docker:
+
+* SLURM (``-profile slurm,singularity``)
```
mberacochea: I would get rid of these profiles (lsf, slurm). Each cluster is different so there is no _universal_ profile for those schedulers

workflows/virify.nf
```diff
@@ -38,51 +37,30 @@ workflow VIRIFY {
...
-                        ["id": id],
-                        [fq1, fq2],
-                    )
-                }
-                else {
-                    exit(1, "input missing, use [--assemble] flag with raw reads")
-                }
-            }
+    groupInputs = { id, assembly, proteins_gff, proteins_faa ->
+        if (params.use_proteins && proteins_gff && proteins_faa) {
```
mberacochea: Do we need the `params.use_proteins`? because if "proteins_gff && proteins_faa" are there that one seems redundant (and maybe even confusing).

mberacochea: If we remove that then we can get rid of this code, and let nf-schema do the heavy lifting

mberacochea (approved):
Excellent @KateSakharova 🥇 . I think we can get rid of a few more things to make it even simpler :)

KateSakharova:
I will fix `use-proteins` in another PR

PR #182, opened 2026-07-17
Fix/remove --use_proteins

KateSakharova:
1. Changing logic of protein prediction. We will run PRODIGAL for all contigs (simulating use-proteins=True) -> separate contigs to HC, LC, PP -> split proteins by categories. Changes:
- filter and run prodigal only for records that do not have GFF and FAA specified
- protein prediction step was moved to main workflow right after LENGTH_FILTERING
- removed coordinates shift function in `write_gff`
"""Adjust CDS coordinates to be relative to the full contig when needed.

    When the pipeline predicted proteins itself (prodigal mode), CDS coordinates in
    the annotation TSV are relative to the prophage region and must be shifted by
    prophage_start to become absolute.  When the user supplied proteins, coordinates
    are already absolute and are returned unchanged.
"""
  - **build_cds_data** no longer shifts CDS start/end by prophage_start at all — that shift only used to happen when use_proteins was False, so now Prodigal always runs on the full assembly (coordinates are always absolute) and use_proteins=True.
  - **get_sequence_regions** always does missed_contigs += 1 + warn on a missing contig

2. Removed `--onlyannotate` option
3. **Replaced** `prodigal` with `pyrodigal`
4. Added **proteins compatability check** for proteins provided in samplesheet. That script returns 3 files:
- matched - proteins are OK in headers in fasta, gff and faa
- not_matched - proteins not matched between those files -> exclude from processing -> add to `not_matched_proteins_report.tsv` file (I haven't done exit 1 yet).
- require_rename - those files match but naming in proteins is truncated to format: `1_1`. It will require the same check and renaming as we do after PYRODIGAL/PRODIGAL


bin/write_viral_gff.py
```diff
@@ -523,10 +477,7 @@ def get_sequence_regions(
         used_contigs.add(clean_contig_name)
         contig_length = contigs_len_dict.get(clean_contig_name)
         if contig_length is None:
-            if use_proteins:
```
mberacochea, line 526: Should we allow this and not scream if a contig is missing?

subworkflows/local/annotate.nf
```diff
@@ -48,22 +47,23 @@ workflow ANNOTATE {
...
-    proteins_fasta = PREDICT_PROTEINS.out.proteins
-    proteins_gff   = PREDICT_PROTEINS.out.proteins_gff
+    
+    category_input
+        .multiMap{ meta, set_name, fasta, faa, gff ->
+            category_fasta: [meta, set_name, fasta]
+            proteins_fasta: [meta, set_name, faa]
+            proteins_gff: [meta, set_name, gff]
+        }
+        .set{input}
```
mberacochea: `.set` is kind of being deprecated by Nextflow, use an assignment instead.

```
input = category_input
          .multiMap{ meta, set_name, fasta, faa, gff -> ...
```

Image

Docs: https://docs.seqera.io/nextflow/reference/operator#set

tests/contig_taxonomic_assign/input_taxonomic_assignment_taxonomy.tsv
mberacochea, line 1: Is this just sorted differently?

bin/write_viral_gff.py
mberacochea: Ignore me.

modules/local/rename_prodigal/main.nf
```diff
@@ -1,19 +1,19 @@
 process RENAME_PRODIGAL {
     label 'process_single'
-    tag "${meta.id} ${confidence_set_name}"
+    tag "${meta.id}"
```
mberacochea, line 3: I like this change, now the joins is cleaner.

mberacochea (changes requested):
Thanks @KateSakharova 
My only comment, is that with the with the removal of the `confidence_set_name` the results for prodigal are going to change.

Docs:
```
    ├── 02-prodigal
    │   ├── high_confidence_viral_contigs_prodigal.faa
    │   └── low_confidence_viral_contigs_prodigal.faa
```
the confidence level is not so the file names will change (right?).

Ales-ibt:
I am reviewing the code, I just want to left a comment here in the meantime. Does this change means that we are not using anymore the user proteins to annotate viphogs? TBH, it makes sense to me, we should control inside the pipeline the gene caller, in particular if we want to replace prodigal by something more suitable for viral genes predictions. But I can still see in the samplesheet json file that we are accepting the user gff and proteins. The question is how are we using those proteins now in the pipeline? And going a step further, how this would this change affects the results in comparison with using the user proteins from the beginning?

KateSakharova:
@Ales-ibt We will use proteins if they are present in samplesheet. Main idea is do not control it just with flag `--use_proteins` - use or not (it should not be binary choice). The main change is about supporting cases like
samplesheet.csv
```
id,assembly,faa,gff
a1,a1.fasta,,
a2,a2.fasta,a2.faa,a2.gff
```
For assembly a1 - prodigal would be run. For assembly a2 - used already presented files

PR #184, opened 2026-08-04
Tests for protein subworkflows

KateSakharova:
That PR should make new protein calling handling more clear.
I've separated **protein calling** and **protein compatibility check** into separate subworkflows and created test for each. That is a new functionality I added in https://github.com/EBI-Metagenomics/emg-viral-pipeline/pull/182. I guess that might be things to change...

**Protein calling:**
pyrodigal + rename of short names. 

I have noticed that sometimes prodigal renamed names in gff `ID=` field to 1_1 or 1_2 and put those names correspondinly to `faa` file. In that case we lost viral identifier (`contig|viral_sequence`) from annotation. That is why rename step is looking for pattern `r"\d+_\d+" ` and renames those back to contig name.
I was not able to reproduce that behaviour with test yet. General test that I added shows that rename step was launched but did nothing.

**Proteins compatibility:**

User can provide proteins as input in samplesheet in faa + gff. We need to check that:
1. contig names in faa correspond to ID= in GFF 
2. contig names correspond to first column in GFF
If any of those rules do not work - we think that proteins are weird and mark those "not_matched" and add it to not_matched_report. That record would be excluded from future pipeline execution.
Another scenario: user provided proteins but they have 1_1 or 2_1 names. We will send those to rename function.

Subworkflow does:
- check of compatibility 
- separate into 'matched', 'not_matched' and 'require_rename'
- rename 'require_rename'
- ignore 'not_matched'


subworkflows/local/protein_prediction/main.nf
```diff
@@ -1,5 +1,5 @@
-include { PYRODIGAL       } from '../../modules/nf-core/pyrodigal/main'  
-include { RENAME_PRODIGAL } from '../../modules/local/rename_prodigal'
+include { PYRODIGAL       } from '../../../modules/nf-core/pyrodigal/main'  
+include { RENAME_PRODIGAL } from '../../../modules/local/rename_prodigal'
```
dezordi, line 2: why the module and process name are `rename prodigal` if here we are using pyrodigal to predict proteins?

dezordi (approved):
Nica Kate, LGTM

subworkflows/local/proteins_compatibility/tests/data/matched.gff
chrisAta, line 1: are the gffs in production ones with a `##FASTA# block? if they are then should include it here just in case it messes things up

subworkflows/local/proteins_compatibility/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    // check proteins compatibility with pipeline expectations
+    CHECK_PROTEINS_COMPATIBILITY(
+        records_with_proteins
+    )
+
+    // reduce the "results/<status>" glob output to a single status name per sample
+    compatibility_status = CHECK_PROTEINS_COMPATIBILITY.out.status
+        .map { meta, results ->
+            def resultsList = results instanceof List ? results : [results]
+            tuple(meta, resultsList[0].name)
```
chrisAta, line 19: what's the logic for always keeping the first file? Might be worth having a comment describing it to make it clearer

subworkflows/local/proteins_compatibility/main.nf
```diff
@@ -0,0 +1,59 @@
+include { CHECK_PROTEINS_COMPATIBILITY      } from '../../../modules/local/check_proteins_compatibility'
+include { RENAME_PRODIGAL                   } from '../../../modules/local/rename_prodigal'
+
+
+workflow PROTEINS_COMPATIBILITY {
+    take:
+    records_with_proteins // [meta, fasta, gff, faa]
```
chrisAta: if the last fasta file is an `faa`, does this mean the first one is a `fna`? maybe update the comment so it's clear

subworkflows/local/proteins_compatibility/main.nf
```diff
@@ -0,0 +1,59 @@
...
+    RENAME_PRODIGAL(
+        checked_records.require_rename.map { meta, _fasta, proteins_gff, proteins_faa -> tuple(meta, proteins_gff, proteins_faa) }
+    )
+    renamed_records = checked_records.require_rename
+        .map { meta, fasta, _proteins_gff, proteins_faa -> tuple(meta, fasta, proteins_faa) }
+        .join(RENAME_PRODIGAL.out.gff)
+        .map { meta, fasta, proteins_faa, renamed_gff -> tuple(meta, fasta, renamed_gff, proteins_faa) }
+
+    emit:
+    checked_records = checked_records.matched
```
chrisAta: should this be renamed to `matched_records`?

chrisAta (approved):
looks good to me! just a few comments

subworkflows/local/proteins_compatibility/tests/data/matched.gff
KateSakharova: I think in virify we don't store fasta in gff. MAP does it for full output

PR #185, opened 2026-08-04
Fix/GitHub actions

KateSakharova:
- Added linting and pre-commit
- Lint files
- Added nf-test for subwfs (because only those have nf-test)
- I'm currently thinking how to create e2e test. Should be easy on MAG that I added

The majority of fixes are just pre-commit lint result


.github/workflows/lint.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: |
+            3.11
+            3.12
```
mberacochea: I would only run this with the one python version

.github/workflows/nf-test-e2e.yml
```diff
@@ -0,0 +1,65 @@
...
+          docker system prune -af || true
+          docker volume prune -f || true
+
+          sudo rm -rf /usr/share/dotnet || true
+          sudo rm -rf /opt/ghc || true
+          sudo rm -rf /usr/local/lib/android || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL || true
+
+          echo "Disk after cleanup:"
+          df -h
```
mberacochea: Use this one instead:
```yml
      - name: Free up disk space
        uses: jlumbroso/free-disk-space@main
        with:
          tool-cache: false
```

.github/workflows/nf-test-subwf.yml
```diff
@@ -0,0 +1,65 @@
...
+          docker system prune -af || true
+          docker volume prune -f || true
+
+          sudo rm -rf /usr/share/dotnet || true
+          sudo rm -rf /opt/ghc || true
+          sudo rm -rf /usr/local/lib/android || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL || true
+
+          echo "Disk after cleanup:"
+          df -h
```
mberacochea: Use this one instead:
```yml
      - name: Free up disk space
        uses: jlumbroso/free-disk-space@main
        with:
          tool-cache: false
```

.github/workflows/nf-test-subwf.yml
```diff
@@ -0,0 +1,65 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
+        with:
+          java-version: '17'
```
mberacochea: I don't think you need this one, the nextflow one should take care of it

.github/workflows/nf-test-subwf.yml
```diff
@@ -0,0 +1,65 @@
...
+          distribution: 'adopt'
+
+      - name: Setup Nextflow latest-edge
```
mberacochea:
```suggestion
      - name: Setup Nextflow
```

.github/workflows/nf-test-subwf.yml
```diff
@@ -0,0 +1,65 @@
...
+          mv .nf-test nf-test-output
+
+      - name: Store the nf-test results if nf-test failed
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: nf-test-out
+          include-hidden-files: true
+          path: |
+            ${{ github.workspace }}/nf-test-output
```
mberacochea: Why not the "nf-core" style one:
```yml
      - name: Run nf-test
        shell: bash
        env:
          NFT_WORKDIR: "~"
        run: |
          nf-test test tests/ \
            --profile=+${{ matrix.profile }} \
            --ci \
            --verbose \
            --tap=test.tap

            # Save the absolute path of the test.tap file to the output
            echo "tap_file_path=$(realpath test.tap)" >> $GITHUB_OUTPUT

      - name: Generate test summary
        if: always()
        shell: bash
        run: |
          # Add header if it doesn't exist (using a token file to track this)
          if [ ! -f ".summary_header" ]; then
            echo "# nf-test results" >> $GITHUB_STEP_SUMMARY
            echo "" >> $GITHUB_STEP_SUMMARY
            echo "| Status | Test Name | Profile |" >> $GITHUB_STEP_SUMMARY
            echo "|:------:|-----------|---------|" >> $GITHUB_STEP_SUMMARY
            touch .summary_header
          fi

          if [ -f test.tap ]; then
            while IFS= read -r line; do
              if [[ $line =~ ^ok ]]; then
                test_name="${line#ok }"
                # Remove the test number from the beginning
                test_name="${test_name#* }"
                echo "| ✅ | ${test_name} | ${{ matrix.profile }} |" >> $GITHUB_STEP_SUMMARY
              elif [[ $line =~ ^not\ ok ]]; then
                test_name="${line#not ok }"
                # Remove the test number from the beginning
                test_name="${test_name#* }"
                echo "| ❌ | ${test_name} | ${{ matrix.profile }} |" >> $GITHUB_STEP_SUMMARY
              fi
            done < test.tap
          else
            echo "| ⚠️ | No test results found | ${{ matrix.profile }} |" >> $GITHUB_STEP_SUMMARY
          fi
```

mberacochea (changes requested):
I've left some tiny comments.

I would also add a .git-blame-ignore-revs ( https://github.com/EBI-Metagenomics/emgapi-v2/blob/main/.git-blame-ignore-revs ) so we leave out this commit out from blames (which could be useful to trace back issues in the future -- by being able to see when a certain file was modified)

PR #188, opened 2026-09-01
Don't rename the taxonomy results -- also remove failOnError on publish

mberacochea:
The taxonomy file results carried the _split_ in the name. This was introduced in the SPLIT_PROTEINS code recently. I've changed back the names of the files, but in the modules.config / publishDir.

I've removed the failOnError: false -> this could lead to runs that generate no results because a configuration error for example. I'm not sure what was the rationale, but if Nextfow can't copy the outputs it should fail.

I've added the serena tool folder to the ignore dir.

### DIFF
--- a/configs/modules.config
+++ b/configs/modules.config
@@ -14,7 +14,6 @@ process {
             [
                 path: { "${params.output}/${meta.id}/${params.finaldir}/annotation/hmmer" },
                 mode: params.publish_dir_mode,
-                failOnError: false,
                 pattern: "*_annotation.tsv"
             ]
         ]
@@ -26,8 +25,8 @@ process {
                 path: { "${params.output}/${meta.id}/${params.taxdir}" },
                 enabled: params.publish_all,
                 mode: params.publish_dir_mode,
-                failOnError: false,
-                pattern: "*_taxonomy.tsv"
+                pattern: "*_taxonomy.tsv",
+                saveAs: { "${set_name}_annotation_taxonomy.tsv" },
             ],
         ]
     }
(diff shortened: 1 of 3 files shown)


configs/modules.config
```diff
@@ -382,8 +356,6 @@ process {
             path: { "${params.output}/${meta.id}/${params.finaldir}/gff" },
             mode: params.publish_dir_mode,
             saveAs: { filename -> filename.equals('versions.yml') ? null : filename },
-            // The GFF may not be generated... (I'm carrying this over as it was done like that in the past.. we need to review)
-            failOnError: false,
```
Ales-ibt, line 386: IMO, a GFF file should be generated in every single run, even with one line only, to avoid this problems at monitoring successfully completion. I know this means in production the generation of many "empty" files, but results and work dirs are removed after running MAP

configs/modules.config
```diff
@@ -395,8 +367,6 @@ process {
             path: { "${params.output}/${meta.id}/${params.finaldir}/gff" },
             mode: params.publish_dir_mode,
             saveAs: { filename -> filename.equals('versions.yml') ? null : filename },
-            // Tthe GFF may not be generated... (I'm carrying this over as it was done like that in the past.. we need to review)
-            failOnError: false,
```
Ales-ibt, line 399: Why do we have this twice?

Ales-ibt (approved):
It looks ok to me. I still think a gff file should be generated in every run, but we can discuss that in the future

configs/modules.config
mberacochea: Do you mean, having an empty GFF file to make it easier for the downstream analyses easier?. I prefer to create an end-of-run file that has a status: no-gff generated or similar. Something like miassembler - https://github.com/EBI-Metagenomics/miassembler#top-level-reports

configs/modules.config
mberacochea: One is the gff and the other one is the index

PR #189, opened 2026-09-03
checkpoints for no cds sequences

Ales-ibt:
This fixes a crash when a predicted viral contig has no proteins. `split_proteins_by_categories.py` exited successfully but skipped writing its per-category GFF whenever nothing matched, leaving `SPLIT_PROTEINS` without a declared output and killing the run. The script now always writes the GFF, and protein-less contigs are discarded at two checkpoints:
1. before the prediction tools, where contigs with no CDS are dropped, and 
2. after prophage prediction, where categories left with no proteins are dropped before annotation. This second checkpoint is necessary because a prophage interval can contain no CDS even on a protein-rich contig, and those intervals do not exist until the predictions have been parsed, i.e. after detection has already run. 

Discarded contigs are recorded in per-step `*_no_proteins.tsv` reports.

A substancial change in the workflow design is that DETECT now depends on the protein branch.

Unit tests added. 

Thanks for reviewing!


docs/output.md
```diff
@@ -75,8 +76,11 @@ In order to have expanded output with more files use `--publish_all` option in p
     │       └── virsorter_metadata.tsv
     ├── 02-protein-prediction [if proteins_faa and proteins_gff were not provided as input]
     │   ├── ACCESSION.faa.gz
-    │   └── ACCESSION.gff.gz
-    │   └── ACCESSION.fna.gz
+    │   ├── ACCESSION.gff.gz
+    │   ├── ACCESSION.fna.gz
+    │   ├── high_confidence_viral_contigs_no_proteins.tsv
```
KateSakharova: The idea of that folder was about returning predicted proteins with pyrodigal (see comment: `if proteins_faa and proteins_gff were not provided as input`).
I feel like it is not correct place for no_proteins.tsv here. Maybe we should return it in `finaldir`, so all results will have those. If `--publish_all` is turned off then only `finaldir` should be published.

bin/split_proteins_by_categories.py
```diff
@@ -66,6 +66,13 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
         required=False,
         default=None,
     )
+    parser.add_argument(
+        "--dropped-report",
```
KateSakharova: `dropped` is not clear name. Maybe remove that option at all and create that file **always** when you have such a case. Just take basename for output name in order not to have collisions for LC, HC and PP

bin/split_proteins_by_categories.py
```diff
@@ -330,9 +362,26 @@ def grep_proteins(self) -> None:
...
+                    )
+                    dropped_contigs.append((contig_record.id, reason))
+
+        # Always write the GFF when asked for one. A category whose contigs carry no proteins
+        # still has to produce the file the SPLIT_PROTEINS process declares as an output;
+        # the empty category is dropped from the channel afterwards, not here.
+        if self.output_gff:
             self._write_gff(gff_by_contig, contig_lengths)
 
+        if self.dropped_report:
```
KateSakharova: change that condition to `if dropped_contigs`

bin/filter_contigs_no_proteins.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        dest="output",
+        help="Output FASTA with the protein-less contigs removed",
+        required=True,
+    )
+    parser.add_argument(
+        "--dropped-report",
```
KateSakharova: rename it to more clear name like no-proteins-report or something .. dropped is very broad and not clear what was dropped (contigs? proteins? report was dropped?)

workflows/virify.nf
```diff
@@ -132,9 +133,20 @@ workflow VIRIFY {
                  .mix(PROTEINS_COMPATIBILITY.out.renamed_records)
                  .mix(newly_predicted)
 
+    // ----------- checkpoint 1: drop contigs with no proteins before the prediction tools
+    // The assembly DETECT consumes uses temporary contig names while the proteins GFF uses
+    // short names, so the rename mapping is passed in to relate the two.
+    FILTER_NO_PROTEINS(
```
KateSakharova, line 139: I guess you do not need to use `filtered_and_renamed_assembly` and `mapfile` here because `assembly_with_short_contignames` should already have names in needed space. 
So, I think you can just take 
```
FILTER_NO_PROTEINS(
   full_input.map { meta, assembly, proteins_gff, _proteins_faa -> [meta, assembly, proteins_gff] }
)
```
It will require some changes in python script. I guess Claude will fix it easily

KateSakharova (changes requested):
Sorry, it took longer than I expected. I tried to come with some useful comments :D

docs/output.md
Ales-ibt: The no proteins file is more for debugging. We are not reporting the contigs shorter than the threshold either. I can fix it to be generated in the workdir but not published

Ales-ibt: I will move the reports to 08-final

bin/filter_contigs_no_proteins.py
Ales-ibt: That option is gone, we are producing the output always using the base name to avoid collisions between quality files

workflows/virify.nf
Ales-ibt: This would be the case for both, user and prodigal proteins?

Ales-ibt: But in the original code DETECT is taking filtered_and_renamed_assembly. This is the with proteins output we are generating here

Ales-ibt: @KateSakharova can you please double check? This is the last bit where you are asking changes

KateSakharova: It is very confusing workflow, let me try to explain.

- `filtered_and_renamed_assembly` operates in `tmp` names
- `assembly_with_short_contignames` operates in `short` names. In case of MAGs it equals to original name because we don't have spaces in names (MGYGXXX_NN). In case of assembly I hope it also should be the same because `short = original.split(' ')[0]`.

we run:
rename -> length filtering -> restore (short names here) -> protein prediction (short names) -> proteins compatability (also short names, check for input proteins and pipeline predicted proteins should be OK because short name is a substring of original name)
in parallel we run DETECT which doesn't care about proteins at all. We run it on **tmp** names because one of detection tools is weird. Then we rename LC, HC and PP fasta to short names each separately.

As far as I understood you used mapfile to map tmp names from `filtered_and_renamed_assembly` to original names to compare with original fasta. I just though you can ignore mapping here and use already renamed back contigs from `assembly_with_short_contignames`

Ales-ibt: Thanks for the explanation, but in the filtering (by contigs with no CDS) step I am taking the length contigs and comparing with original GFF, that's why I need mapping to original names, at least when user is providing both files

KateSakharova (approved):

